PR #243, opened 2021-05-18
ci: add compose app and dockerifle for licensed

skottler:


docker/Dockerfile.licensed
```diff
@@ -0,0 +1,9 @@
+FROM ruby:2.7-slim-buster
+
+WORKDIR /app
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y cmake pkg-config git-core libgit2-dev
+RUN gem install licensed
+
+ENTRYPOINT ["licensed"]
```
fatih, line 9: I think this needs to run the command `licensed status`, assuming this is the CLI command?

fatih (approved):

PR #246, opened 2021-05-18
Update README.md to include mysql-client install

edgarlanting:
Upon invoking the pscale cli, an error occurs stating:
`Error: couldn't find the 'msyql' client required to run this command.`
To resolve this, you need to run:
`brew install mysql-client` and set the appropriate PATH to be able to use it.
Once this has been done, the Planetscale CLI works properly.

Environment:
M1 ARM Mac
MacOS 11.2.3
Homebrew 3.1.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ PlanetScale is more than a database and our CLI is more than a jumble of command
 ```
 brew install planetscale/tap/pscale
 ```
-!! Note of importance: `pscale` requires the MySQL Client. You can install it by running:
+Optional: `pscale` requires the MySQL Client for certain commands. You can install it by running:
 
 ```
 brew install mysql-client
@@ -37,4 +37,3 @@ brew upgrade pscale
 ### Manually
 
 Download the pre-compiled binaries from the [releases](https://github.com/planetscale/cli/releases/latest) page and copy to the desired location.
-


fatih:
@edgarlanting does it happen on every command or only on certain commands?

edgarlanting:
> @edgarlanting does it happen on every command or only on certain commands?

Hi @fatih amongst others in my case it happened with `pscale shell <dbname> main` when opening a shell to  a db. When for instance running `pscale auth login` it does not happen`

fatih:
> So Imho it's only the shell functionality that requires it.

Great, I just wanted to confirm. What you're describing is correct.

README.md
```diff
@@ -14,6 +14,11 @@ PlanetScale is more than a database and our CLI is more than a jumble of command
 ```
 brew install planetscale/tap/pscale
 ```
+!! Note of importance: `pscale` requires the MySQL Client. You can install it by running:
```
fatih, line 17: We only require it for the `pscale shell` subcommand, otherwise, it's optional. I wonder if should add this bit in paranthese? 

```suggestion
Optional: `pscale` requires the MySQL Client for certain commands. You can install it by running:
```

edgarlanting: As a user, it would make perfect sense to me so I'm keeping my eyes peeled to see what you decide.

fatih: Thank you Edgar.

fatih (approved):

README.md
edgarlanting: Your welcome Fatih, its an awesome db platform😄

croaky:
Should this be `brew install mysql` instead of `mysql-client`?

This was my experience:

```
% pscale shell example main  
Error: couldn't find the 'msyql' client required to run this command.
To install, run: brew install mysql-client
% brew install mysql-client
... (elided for brevity)
==> Summary
🍺  /opt/homebrew/Cellar/mysql-client/8.0.23: 135 files, 151MB
% pscale shell example main  
Error: couldn't find the 'msyql' client required to run this command.
To install, run: brew install mysql-client
```

But, after I `brew install mysql`, it worked. I'm also on an M1 ARM Mac.

edgarlanting:
It should just be the mysql-client. Unless of course you plan to do more with MySQL

fatih:
The `mysql` package also includes the server and the DB itself. You don't need it.

croaky:
Thanks! Missing `PATH="$BREW/opt/mysql-client/bin:$PATH"` where

```
if [ "$arch" = "arm64" ]; then
  BREW="/opt/homebrew"
else
  BREW="/usr/local"
fi
```

fatih:
Heads up that I've created a PR to improve the situation. If you have any comments or feedback let me know please: https://github.com/planetscale/cli/pull/256

croaky:
@fatih That looks great. What do you think about one more check for arm64? https://github.com/planetscale/cli/pull/258

edgarlanting:
Thanks @fatih

PR #256, opened 2021-05-19
cmdutil: search `mysql` under the Homebrew preferred location

fatih:
This PR makes sure to search under the `/usr/local/opt/mysql-client/bin/` path for the `mysql` client. The `mysql-client` package installs the client in an unusual location. It's not part of `PATH`; hence we can't find it and prompt the user to install it again.


mscoutermarsh (approved):

marwan-at-work (approved):
Thank you <3

PR #260, opened 2021-05-20
connect.go: let the user configure the protocol for DATABASE_URL

nickvanw:
This gives the user some configuration over what protocol gets dropped into their `DATABASE_URL`. Some applications want `mysql`, and `mysql2` is often used in Rails. This defaults to what is there today, but allows the user to override it if they want. 

cc @iheanyi

### DIFF
--- a/internal/cmd/connect/connect.go
+++ b/internal/cmd/connect/connect.go
@@ -24,10 +24,11 @@ import (
 
 func ConnectCmd(ch *cmdutil.Helper) *cobra.Command {
 	var flags struct {
-		port        string
-		host        string
-		remoteAddr  string
-		execCommand string
+		port                string
+		host                string
+		remoteAddr          string
+		execCommand         string
+		execCommandProtocol string
 	}
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


iheanyi (approved):
LGTM. Should we validate that it's either `mysql` or `mysql2`?

internal/cmd/connect/connect.go
```diff
@@ -122,6 +123,7 @@ argument:
 		"PlanetScale Database remote network address. By default the remote address is populated automatically from the PlanetScale API.")
 	cmd.MarkPersistentFlagRequired("org") // nolint:errcheck
 	cmd.PersistentFlags().StringVar(&flags.execCommand, "execute", "", "Run this command after successfully connecting to the database.")
+	cmd.PersistentFlags().StringVar(&flags.execCommandProtocol, "execute.protocol", "mysql2", "Protocol for the DATABASE_URL value in execute")
```
fatih, line 126: We use the dash chracter for separating words. Let's make sure we're consistent with flag formats. 

```suggestion
	cmd.PersistentFlags().StringVar(&flags.execCommandProtocol, "execute-protocol", "mysql2", "Protocol for the DATABASE_URL value in execute")
```

fatih (approved):

Issue #265, opened 2021-05-21
[Windows] Error: error net.Listen: listen tcp 127.0.0.1:3306: bind: An attempt was made to access a socket in a way forbidden by its access permissions.


janpio:
The Windows CLI throws the following error when port 3306 is already in use:

```
C:\Users\Jan\Documents\throwaway\planetscale  (planetscale@1.0.0)
λ pscale connect fk-test foo
Error: error net.Listen: listen tcp 127.0.0.1:3306: bind: An attempt was made to access a socket in a way forbidden by its access permissions.
```

You could either have a more descriptive error message, or automatically use a different port that is available.

fatih:
Thank you @janpio for the issue. We're going to have a look at it.

For the record, we actually try a random port, but only if the port is in use. The operating system returns a custom error that we catch. See: https://github.com/planetscale/cli/blob/1bd7252ab047c9f3029217e210c4b847285b14f5/internal/cmd/connect/connect.go#L197

It looks like the error message you're seeing is different. We don't check for that specific error message.

janpio:
Ok. I had MySQL 5.7 running on that port via the default install, managed via MySQL Workbench (Where I then could not really stop the process/service and had to kill it manually via a Admin Taskmgr). Maybe that info helps you to pin it down.

fatih:
Hi @janpio, 

Thanks again for the feedback. I opened a [PR](https://github.com/planetscale/cli/pull/270) with a potential fix. You can try it out by checking out my branch and compiling it yourself. Here is one way to use the [gh](https://github.com/cli/cli#installation) tool. Run the commands from the root of the Git repository:

```
gh pr checkout 270
cd cmd/pscale
go build 
```

Alternatively, I also compiled it from my branch, if you would like to test it out:

[pscale.exe.zip](https://github.com/planetscale/cli/files/6531420/pscale.exe.zip)

janpio:
Still have this tab open, and will get to it sooner or later to reproduce this once again and confirm the fix.

PR #270, opened 2021-05-24
cmd/connect: handle Windows socket error 10013 (WSAEACCES)

fatih:
This PR also handles the case when a socket fails to bind to a socket with exclusive access. For more information, check out the docs for the code `10013`: https://docs.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2

fixes: https://github.com/planetscale/cli/issues/265


iheanyi (approved):
Nice!

PR #278, opened 2021-05-25
connect: exit if `--execute` finishes

fatih:
This PR changes the behavior of the `--execute` flag.

Previously we would continue to serve the proxy even if the command with `--execute` would finish or fail. With this PR, we're changing it.  Instead of silently continuing to run, we're exiting entirely and shutting down the proxy. This is the correct behavior because the user explicitly has passed the `--execute` flag, and we should assume they only want to run the proxy for a single command.

If there are use cases where we don't want to exit, it's also easy to add it now. We could introduce a new flag that changes whether to call `cancel()` or not.

### DIFF
--- a/internal/cmd/connect/connect.go
+++ b/internal/cmd/connect/connect.go
@@ -105,7 +105,7 @@ argument:
 				go func() {
 					err := runCommand(ctx, flags.execCommand, flags.execCommandProtocol, database, branch, proxyReady)
 					if err != nil {
-						ch.Printer.Println("running command with --execute has failed: %s", err)
+						ch.Printer.Printf("running command with --execute has failed: %s", err)
 					}
 
 					// TODO(fatih): is it worth to making cancellation configurable?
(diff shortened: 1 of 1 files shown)


internal/cmd/connect/connect.go
```diff
@@ -99,15 +102,23 @@ argument:
...
 
 			if flags.execCommand != "" {
-				go runCommand(ctx, flags.execCommand, flags.execCommandProtocol, database, branch, proxyReady)
+				go func() {
+					err := runCommand(ctx, flags.execCommand, flags.execCommandProtocol, database, branch, proxyReady)
+					if err != nil {
+						ch.Printer.Println("running command with --execute has failed: %s", err)
+					}
+
+					// TODO(fatih): is it worth to making cancellation configurable?
```
iheanyi, line 111: Yes, I think so.

iheanyi (approved):

nickvanw (approved):

Issue #254, opened 2021-05-18
Issues with the current deployment process


baetheus:
The current production deployment process advises wrapping an existing application or executable in the pscale client. This becomes problematic for the following reasons:

1. Not all existing apps are keyed to look for the `DATABASE_URL` env variable in their configurations.
2. Not all existing apps are directly managed (some are packaged upstream and are pulled into the same pod as the application) and so require repackaging to wrap in the pscale binary.
3. For larger scale deployments where processing cycles or space are in short supply, running an extra 9Mb process alongside each application could be a non-starter.
4. For other, uptime sensitive deployments, running the primary application as a sub-process of pscale could be a non-starter.

It is with these problems in mind that I request a standalone container image blessed by planetscale that takes the following env vars:

```
PLANETSCALE_ORG: <ORG_NAME>
PLANETSCALE_SERVICE_TOKEN_NAME: <SERVICE_TOKEN_NAME>
PLANETSCALE_SERVICE_TOKEN: <SERVICE_TOKEN>
PLANETSCALE_PORT: <EXPOSED_PORT>
PLANETSCALE_DATABASE: <DATABASE_NAME>
PLANETSCALE_BRANCH: <DATABASE_BRANCH>
```

And exposes the resulting forwarding service on `PLANESCALE_PORT`. This would solve all four of the problems outlined above and is in fact what I ended up deploying to test planetscale. I hope this helps, I'm looking forward to playing with your product.

nickvanw:
Hey @baetheus!

This lines up with something that we've been looking to produce indeed - the [sidecar pattern](https://docs.microsoft.com/en-us/azure/architecture/patterns/sidecar) is very common, and something that PlanetScale's CLI can easily do inside of a separate container from an application. 

I can't give an exact ETA, but this is high up on the list of libraries and tools that we want to provide to make it easier for people to use PlanetScale in their applications.

fatih:
Hi folks. I created a simple PR to kickstart the conversation: https://github.com/planetscale/cli/pull/280

`pscale connect` will probably only work if you expose 3306 to host 3306. I've added more details on the PR itself for our upcoming plans. We might expose the sidecar container as part of the https://github.com/planetscale/sql-proxy repository. 


> And exposes the resulting forwarding service on PLANESCALE_PORT

Can `docker` expose a port of a container to the host it's running? Or do you run your application in Kubernetes or similar environments?

baetheus:
[...]

PR #280, opened 2021-05-25
Add Dockerfile to run `pscale` as a container

fatih:
This PR introduces a simple `Dockerfile` that allows us to run `pscale` from within a container. Here are the commands if you want to try it out on your localhost:

```shell
# build the image
docker build -t pscale:latest .

# run the container
# create the service tokens via 'pscale service-token create'
docker run -it --env PLANETSCALE_SERVICE_TOKEN= <token> --env PLANETSCALE_SERVICE_TOKEN_NAME=<name> --env PLANETSCALE_ORG=<org>  pscale shell planetscale main
``` 

Three environment variables are required to make it work:

```
PLANETSCALE_SERVICE_TOKEN
PLANETSCALE_SERVICE_TOKEN_NAME
PLANETSCALE_ORG
```

Users can also use the `--env-file` option (https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file) to set the environment variables.

 There are several things that I'm planning to do in follow-up PRs: 

* Push to a public registry
* Configure `goreleaser` so it automatically publishes with the CLI version
* Check `pscale connect` and make it work. Users who wish to use `pscale connect` as a sidecar pattern will probably provide a container based on https://github.com/planetscale/sql-proxy. The `pscale connect` functionality is built on top of `planetscale/sql-proxy` codebase.

/xref https://github.com/planetscale/cli/issues/254

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,7 @@ COPY . .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build github.com/planetscale/cli/cmd/pscale
 
 FROM alpine:latest  
-RUN apk --no-cache add ca-certificates
-RUN apk add mysql-client
+RUN apk --no-cache add ca-certificates mysql-client
 
 WORKDIR /app
 COPY --from=build /app/pscale /usr/bin
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.16 as build
+WORKDIR /app
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build github.com/planetscale/cli/cmd/pscale
+
+FROM alpine:latest  
+RUN apk --no-cache add ca-certificates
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.16 as build
+WORKDIR /app
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build github.com/planetscale/cli/cmd/pscale
+
+FROM alpine:latest  
+RUN apk --no-cache add ca-certificates
+RUN apk add mysql-client
```
nickvanw, line 8:
```suggestion
RUN apk --no-cache add ca-certificates mysql-client
```

should we do this all in one `RUN` command?

nickvanw (approved):

Dockerfile
fatih: Yeap, makes sense 👍🏼

Issue #283, opened 2021-05-25
connect: bubble up exit code of `--execute` command when done


janpio:
0.40.0 added the great feature that `connect` ends when the command in `--execute` ends. Unfortunately `pscale connect` still exits with status code 1, which a test runner or CI system of course interpret as a failure: https://github.com/prisma/e2e-tests/pull/1700/checks?check_run_id=2669053383#step:5:134

It would be great if the CLI could pick up the exit code of the wrapped command, and bubble that up and exit `pscale connect` with that.

fatih:
Hi @janpio 

You're right; it shouldn't exit with `1`. I checked the logs above, though, and it looks like the proxy (`pscale connect`) left with no issues. In that case, we exit with the status `0`. Did you see any other logs that could indicate an error? I wonder if we somehow return an error in the case where we should log it.

I'm currently looking into wrapping the execute command's status code and will share a working branch with you.

janpio:
> You're right; it shouldn't exit with `1`. I checked the logs above, though, and it looks like the proxy (`pscale connect`) left with no issues. In that case, we exit with the status `0`. Did you see any other logs that could indicate an error? I wonder if we somehow return an error in the case where we should log it.

I based this off `pscale connect e2e-tests main --execute-protocol 'mysql' --debug --execute 'yarn test'` being the last command being run, and then the process exiting with status code 1:
```
exitting with code 1
Warning: Attempt 1 failed. Reason: Child_process exited with error code 1
```

I can retry to get a manual reproduction.

fatih:
@janpio This issue should be now fixed with the PR: https://github.com/planetscale/cli/pull/287

I released a new version with this improvement. If you update to `v0.42.0` (https://github.com/planetscale/cli/releases/tag/v0.42.0), you should be able to try it out. Please let me know if you see any issues.

janpio:
Thanks!

Somehow it is not working for me/us:

```
run.sh was successful, running databases/planetscale/test.sh...

pscale version 0.42.0 (build date: 2021-06-01T14:28:46Z commit: 59880fe)

2021-06-01T14:48:17Z	info	adding tls.Config to the cache	{"instance": "***/e2e-tests/main"}
2021-06-01T14:48:17Z	info	ready for new connections
2021-06-01T14:48:17Z	info	listening remote DB instance	{"local_addr": "127.0.0.1:3306", "instance": "***/e2e-tests/main"}
Secure connection to database e2e-tests and branch main is established!.

[...]

PR #288, opened 2021-06-02
cmd: improve execution of the root command

fatih:
This PR simplifies the execution and printing of the error message of the root command. Previously, we would print an error message inside the `cmd/pscale` package and the `internal/cmd` package. The exit status returned from the root command would also be shadowed from another function call. In our case, this was the `updater` logic. Having these two operations in different places makes it difficult to debug and also introduces bugs (see: https://github.com/planetscale/cli/issues/283#issuecomment-852393281)

With this PR, the `cmd/pscale` main package only exits with a given exit code. It's as minimal as its gets. By making sure we're also just exiting with a code in the main package, we also set an artificial boundary. This way, we don't print any message accidentally inside the `main` package (because all it receives is an integer value).

Next, we separated the actual execution of the root command and any post-function calls. The root command is now called with the `cmd.runCmd()` function. The `cmd.Execute()` is now responsible for calling the root command, printing any useful information/errors to the users, and returning the **correct** exit code to the `main` package. 

closes: https://github.com/planetscale/cli/issues/283

### DIFF
--- a/cmd/pscale/main.go
+++ b/cmd/pscale/main.go
@@ -1,13 +1,9 @@
 package main
 
 import (
-	"errors"
-	"fmt"
 	"os"
 
 	"github.com/planetscale/cli/internal/cmd"
-	"github.com/planetscale/cli/internal/cmdutil"
-	"github.com/planetscale/cli/internal/printer"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


iheanyi (approved):
LGTM!

PR #306, opened 2021-06-11
Add region flag to database and branch creation commands

iheanyi:
This pull request adds the `--region` flag to the `branch create` and `database create` commands so we can eventually make regional databases and branches as we add them in. Also, we can get some auto-completion action going with hitting the regions API to pre-populate things.

Depends on https://github.com/planetscale/planetscale-go/pull/70 landing first and we gotta update our `planetscale-go` version.

### DIFF
--- a/internal/cmd/branch/create.go
+++ b/internal/cmd/branch/create.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/browser"
 	"github.com/planetscale/cli/internal/cmdutil"
 	"github.com/planetscale/cli/internal/printer"
+	"github.com/planetscale/planetscale-go/planetscale"
 	ps "github.com/planetscale/planetscale-go/planetscale"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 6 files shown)


internal/cmd/branch/create.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/browser"
 	"github.com/planetscale/cli/internal/cmdutil"
 	"github.com/planetscale/cli/internal/printer"
+	"github.com/planetscale/planetscale-go/planetscale"
```
mattrobenolt, line 11: Isn't this already being imported in the line below aliased as `ps` ?

iheanyi: Good catch, thanks! Fixed in https://github.com/planetscale/cli/pull/306/commits/afe9e35433d292431d3669cfebd0c1a2bcc61636.

internal/cmd/database/create.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/planetscale/cli/internal/cmdutil"
 	"github.com/planetscale/cli/internal/printer"
 
+	"github.com/planetscale/planetscale-go/planetscale"
```
mattrobenolt, line 11: You duped the import here too. :)

iheanyi: 🤦🏾

iheanyi: Fixed (again) in https://github.com/planetscale/cli/pull/306/commits/23ead1fbeabfcc71820e043f8cd058146e9970e5

fatih (approved):
Looks good to me 👍🏼

iheanyi:
Marked the region flags as hidden for now, but they are implemented.

PR #309, opened 2021-06-15
completion: include completions in archives, linux packages and homebrew

fatih:
This PR includes the completions script generated by `pscale completion`
and adds them for the macOS and Linux packages. That way, the user
doesn't have to generate and install the completion scripts manually.


mscoutermarsh (approved):
🤘

PR #325, opened 2021-07-06
cli: add context cancellation support to all commands

fatih:
This PR adds proper context cancellation and propagation support to all commands. Even if the user cancels the operation, we will continue to make requests to our API.

With the changes introduced in this PR, we now listen to `os.Interrupt` and `os.Kill` signals. The context associated with the signals is passed down to all child commands. Any command accepting a `ctx` value will now be canceled automatically. 

Another benefit of this PR is that we got rid of custom context implementations (such as `pscale connect` or `pscale shell`). Now, all commands use the same context value created in the root command.

### DIFF
--- a/cmd/pscale/main.go
+++ b/cmd/pscale/main.go
@@ -15,8 +15,12 @@ var (
 )
 
 func main() {
+	os.Exit(realMain())
+}
+
+func realMain() int {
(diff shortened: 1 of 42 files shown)


iheanyi (approved):
Yes! Good call on this!

PR #329, opened 2021-07-09
Switch to ECDSA for client certificate

dbussink:
These are smaller & faster so let's prefer ECDSA.


fatih (approved):
Looks good to me 👍🏼

PR #331, opened 2021-07-09
ci: pass down the release version to the next step

fatih:


.buildkite/pipeline.yml
```diff
@@ -19,20 +19,22 @@ steps:
...
 - wait
 
 - label: ":github: Publishing artifacts"
   if: build.branch == "main"
+  command: script/release.sh
   plugins:
     - docker#v3.8.0:
-        image: "goreleaser/goreleaser:v0.169.0"
-        command: ["release", "--rm-dist"]
+        image: "golang:1.16.5"
```
dbussink, line 33: Is there a way to pull this out into a variable somewhere for when we want to update Go?

fatih: We could expose that from the agents in the BuildKite UI as an environment variable and use interpolation for this field. 

Or, we could create a dynamic pipeline that reads this value from somewhere and then updates all scripts/yaml files.

dbussink: @fatih Yeah, I don't really have a strong opinion on how, but it would be nice to not have to update too many places somewhere deep in configuration :smile:.

fatih: I'm more than happy to improve the release pipeline and our scripts once the release process is up and running :)

dbussink: Sound good :+1:.

dbussink (approved):

PR #337, opened 2021-07-13
errors: show the user a helpful message if their token has expired

fatih:
This PR tries to interpret an authentication error message. The API can return an error message in the form:

```json
{
  "error":"invalid_token",
  "error_description":"The access token is invalid",
  "state":"unauthorized"
}
```

which is of non-standard type. To handle it properly, we're trying to unmarshal the error message into the type:

```go
// authErrorResponse represents an error response from the API
type authErrorResponse struct {
	Error            string `json:"error"`
	ErrorDescription string `json:"error_description"`
	State            string `json:"state"`
}
```

And then check whether the `Error` field is of type `invalid_token`, which is an indication of expired tokens.

### DIFF
--- a/internal/cmdutil/errors.go
+++ b/internal/cmdutil/errors.go
@@ -1,6 +1,8 @@
 package cmdutil
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/planetscale/planetscale-go/planetscale"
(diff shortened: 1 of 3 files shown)


internal/cmdutil/errors.go
```diff
@@ -43,6 +45,26 @@ func HandleError(err error) error {
...
+
+		errorRes := &authErrorResponse{}
+		mErr := json.Unmarshal([]byte(err.Error()), errorRes)
+		if mErr != nil {
+			// return back original error (not *mErr*). Looks like the error is
+			// not an authentication error
+			return err
+		}
+
+		if errorRes.Error == "invalid_token" {
```
dbussink, line 64: Is an invalid token also due to an expired token? Are there other cases where the token is returned as invalid?

fatih: It actually has three states, but that's not exposed in the API:  

``` 
invalid_token:
  revoked: "The access token was revoked"
  expired: "The access token expired"
  unknown: "The access token is invalid"
``` 

This is the error body we receive:

```
{
  "error":"invalid_token",
  "error_description":"The access token is invalid",
  "state":"unauthorized"
}
```

I couldn't find the place where we return this error message though. I think Rails handles it implicitly. I'll have to sync with the team later in my evening to understand how it decides what to return.

iheanyi (approved):

nickvanw (approved):

mdlayher (approved):
LGTM, possible to add some coverage on this new path?

I see this has been outstanding for a bit, so I figured I would give it a review to see if this is something we'd still like to pursue. Thanks!

PR #339, opened 2021-07-14
Update to latest planetscale-go and sql-proxy

dbussink:
This ensures we can use the multi CA support and that that is all working now.


fatih (approved):
Looks good to me 👍🏼

PR #344, opened 2021-07-19
Update to the latest planetscale go client

no-itsbackpack:
This fixes the bugs we are having with users creating service-tokens
```go
~$ pscale service-token add-access te2okcpjdb5t connect_branch --database again
Error: json: cannot unmarshal string into Go struct field ServiceTokenAccess.data.id of type int
~$ pscale service-token show-access te2okcpjdb5t
Error: json: cannot unmarshal string into Go struct field ServiceTokenAccess.data.id of type int
```


nickvanw (approved):

iheanyi (approved):
You might want to run `go mod tidy` before landing this! :D

PR #346, opened 2021-07-20
cmd: add `pscale audit-log` command

fatih:
This PR adds the new `pscale audit-log` command to list audit logs. The user can filter out audit logs for a specific event type by specifying the `--action` flag several times.


iheanyi (approved):
Nice!

PR #359, opened 2021-08-04
Cleanup usage of database credentials

dbussink:
These values are hardcoded right now and are not actually providing anything useful. The local proxy always uses root with no password anyway, so we can depend on that.


dbussink:
This also ensures things keep working when planetscale-go is updated with https://github.com/planetscale/planetscale-go/pull/82.

fatih (approved):
SGTM 👍🏼

PR #362, opened 2021-08-11
Add `branch promote` command

iheanyi:
This pull request adds the `branch promote` command from promoting a development branch to a production branch. 

Also did some cleanup here and there.


internal/cmd/branch/create.go
```diff
@@ -57,11 +57,6 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 			source := args[0]
 			branch := args[1]
 
-			// Simplest case, the names are equivalent
-			if source == branch {
-				return fmt.Errorf("a branch named '%s' already exists", branch)
-			}
-
```
fatih, line 64: Was this removed intentionally?

fatih (approved):
Looks good to me 👍🏼 Just added a question regarding some removed lines.

internal/cmd/branch/create.go
iheanyi: Yep!

PR #370, opened 2021-08-17
passwords: support listing all passwords

fatih:
This PR changes the `pscale password list` subcommand to list all branches. The following changes were made:

* Renamed the `passwords` package to `package` (packages in Go should be singular)
* Added a new header with the name `Branch` to the output, so the user can see which password belongs to which branch
* `pscale password list database` now lists all passwords from all branches. This is now the same behavior we see in the UI
* `pscale password list database branch` lists all passwords for the given `branch`.

Depends on https://github.com/planetscale/cli/pull/367


Phanatic (approved):
Thanks for doing this!

PR #371, opened 2021-08-17
restore: set workload=olap

nickvanw:
We do this in the dump command (https://github.com/planetscale/cli/blob/12faa1658517b2677877c9cf5821c1550c7b2c3f/internal/cmd/database/dump.go#L143), I think we probably want to in the restore command as well. 

This should make imports more smooth for more users, I believe.


skottler (approved):
🙇

deepthi:
I don't think you can insert in olap mode.

PR #375, opened 2021-08-18
Add override to allow non-interactive use of `pscale shell`

Sinjo:
This makes it possible to use `pscale shell` to run small SQL scripts
against a database branch from shell scripts.

This is behind an environment variable for now, while we make sure there
aren't unintended consequences. We may remove the check entirely at a
later time.

### DIFF
--- a/internal/cmd/shell/shell.go
+++ b/internal/cmd/shell/shell.go
@@ -54,7 +54,9 @@ second argument:
 			database := args[0]
 
 			if !printer.IsTTY || ch.Printer.Format() != printer.Human {
-				return errors.New("pscale shell only works in interactive mode")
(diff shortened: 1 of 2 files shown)


internal/cmd/shell/shell.go
```diff
@@ -54,7 +54,9 @@ second argument:
 			database := args[0]
 
 			if !printer.IsTTY || ch.Printer.Format() != printer.Human {
-				return errors.New("pscale shell only works in interactive mode")
+				if !(os.Getenv("PSCALE_ALLOW_NONINTERACTIVE_SHELL") == "true") {
```
fatih, line 57: I think we should just check whether it exists or not, instead of checking the value:

```suggestion
				if os.Getenv("PSCALE_ALLOW_NONINTERACTIVE_SHELL") == "" {
```

fatih (approved):

PR #386, opened 2021-09-07
Fix for Windows colors

mattn:
before
image
after
image

### DIFF
--- a/internal/cmd/connect/connect.go
+++ b/internal/cmd/connect/connect.go
@@ -17,6 +17,7 @@ import (
 	"github.com/planetscale/cli/internal/proxyutil"
 	"github.com/planetscale/planetscale-go/planetscale"
 
+	"github.com/fatih/color"
 	"github.com/mattn/go-shellwords"
 	"github.com/planetscale/sql-proxy/proxy"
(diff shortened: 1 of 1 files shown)


nickvanw (approved):

iheanyi:
Thanks for the PR!

internal/cmd/connect/connect.go
```diff
@@ -173,7 +175,7 @@ func runProxy(ctx context.Context, proxyOpts proxy.Options, database, branch str
 			return
 		}
 
-		fmt.Printf("Secure connection to database %s and branch %s is established!.\n\nLocal address to connect your application: %s (press ctrl-c to quit)\n",
+		fmt.Fprintf(color.Output, "Secure connection to database %s and branch %s is established!.\n\nLocal address to connect your application: %s (press ctrl-c to quit)\n",
```
fatih, line 178: We should use ch.Printf which handles this. Looks like we forget to change it here. Thanks for the change though 👍

jasonlong:
Is there anything that can be done to improve this blue? It's 100% unreadable to me in that screenshot. 😬

PR #398, opened 2021-09-14
Use new branch promotion APIs for `promote` command

iheanyi:
This pull request reworks the `promote` function to use the new branch-level API functions that we implemented in the Go SDK. Given the new async nature of this endpoint, we first promote and then poll until the branch promotion process has either completed, errored, or timed out.

This pull request depends on https://github.com/planetscale/planetscale-go/pull/97.


internal/cmd/branch/promote.go
```diff
@@ -78,13 +79,46 @@ func PromoteCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				if err != nil {
+					switch cmdutil.ErrCode(err) {
+					case ps.ErrNotFound:
+						return fmt.Errorf("promotion request for branch %s does not exist in database %s", printer.BoldBlue(branch), printer.BoldBlue(source))
+					default:
+						return cmdutil.HandleError(err)
+					}
+				}
+
+				time.Sleep(1 * time.Second)
```
fatih, line 99: We should have a timeout, like 4 or 5 minutes and bailout. Otherwise, in case of a wrong API, this will loop forever. For that, I would refactor this for loop into a function to make it easy, and use a `for/select` idiom with two tickers. On being a one-second ticker that checks the state, and the other select channel will be `time.After(timeout)` and return an error.  Here is an example:

```go
func waitPromoteState() (bool, error) {
    timeout := time.After(5* time.Minute)
    ticker := time.NewTicker(time.Second)
    defer ticker.Stop()


    for {
        select {
        case <-timeout:
            return false, errors.New("timed out")
        case <-ticker.C:
            // check promote status 
        }
}
```

fatih (approved):
LGTM. Added a comment about timing out, I think once you refactor that, we're good to go. In the future, we can replace that with exponential backoff, but I think it's an overkill for now, so the one-second ticker is all good.

PR #422, opened 2021-10-31
Provide Linux examples how to install pscale

jonico:
* provide concrete, scriptable examples how to install pscale on RPM and deb- based Linux distributions


fatih:
@jonico thank you for the contribution. I think the current instructions are good enough. We don't want to populate the readme and make it hard to read. We also have instructions here: https://docs.planetscale.com/reference/planetscale-environment-setup#linux-instructions I'm closing this PR, but if you want you can add a reference to the URL here.

PR #430, opened 2021-11-28
Add M1 Mac homebrew install fallback docs link

borjadotai:
I have been struggling to install the CLI, especially through Homebrew. I found a StackOverflow that explains the steps to be able to install it through Rosetta which was the only way I could get it to work. Thought it would be worth adding a little footnote to the step as a fallback if someone else also runs into this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ PlanetScale is more than a database and our CLI is more than a jumble of command
 ```
 brew install planetscale/tap/pscale
 ```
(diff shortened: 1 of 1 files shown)


dbussink:
@borjadotai Is this one still an issue? I have an M1 machine and had no issues with the install and we do publish ARM64 builds for MacOS these days. So I'm inclined to close this PR at this point.

PR #433, opened 2021-11-29
Add `auth refresh` command

iheanyi:
This pull request adds a command for adding refreshing of expired authentication using `pscale auth refresh` vs. re-running `pscale auth login`. Don't necessarily have to merge this, but it's something to consider.

### DIFF
--- a/internal/auth/authenticator.go
+++ b/internal/auth/authenticator.go
@@ -232,6 +232,35 @@ func (d *DeviceAuthenticator) requestToken(ctx context.Context, deviceCode strin
 	return tokenRes.AccessToken, tokenRes.RefreshToken, nil
 }
 
+// RefreshToken for an access token
+func (d *DeviceAuthenticator) RefreshToken(ctx context.Context, token string) (string, string, error) {
+	payload := strings.NewReader(fmt.Sprintf("client_id=%s&client_secret=%s&refresh_token=%s&grant_type=refresh_token", d.ClientID, d.ClientSecret, token))
+	req, err := d.NewFormRequest(ctx, http.MethodPost, "oauth/token", payload)
+	if err != nil {
+		return "", "", errors.Wrap(err, "error creating request")
+	}
+
+	res, err := d.client.Do(req)
+	if err != nil {
+		return "", "", errors.Wrap(err, "error performing http request")
+	}
+
+	defer res.Body.Close()
+
+	tokenRes := &OAuthTokenResponse{}
+
+	err = json.NewDecoder(res.Body).Decode(tokenRes)
+	if err != nil {
+		return "", "", errors.Wrap(err, "error decoding token response")
+	}
+
+	if tokenRes.AccessToken == "" && tokenRes.RefreshToken == "" {
+		return "", "", errors.New("error refreshing tokens, both access token and refresh token are blank")
+	}
+
+	return tokenRes.AccessToken, tokenRes.RefreshToken, nil
+}
+
 // RevokeToken revokes an access token.
 func (d *DeviceAuthenticator) RevokeToken(ctx context.Context, token string) error {
 	payload := strings.NewReader(fmt.Sprintf("client_id=%s&client_secret=%s&token=%s", d.ClientID, d.ClientSecret, token))
--- a/internal/auth/authenticator.go
+++ b/internal/auth/authenticator.go
@@ -30,7 +30,7 @@ const (
 // Authenticator is the interface for authentication via device oauth
 type Authenticator interface {
 	VerifyDevice(ctx context.Context) (*DeviceVerification, error)
-	GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (string, error)
+	GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (string, string, error)
 	RevokeToken(ctx context.Context, token string) error
 }
 
@@ -172,13 +172,10 @@ func (d *DeviceAuthenticator) VerifyDevice(ctx context.Context) (*DeviceVerifica
 
 // GetAccessTokenForDevice uses the device verification response to fetch an
 // access token.
-func (d *DeviceAuthenticator) GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (string, error) {
-	var accessToken string
-	var err error
-
+func (d *DeviceAuthenticator) GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (accessToken string, refreshToken string, err error) {
 	for {
 		time.Sleep(v.CheckInterval)
-		accessToken, err = d.requestToken(ctx, v.DeviceCode, d.ClientID)
+		accessToken, refreshToken, err = d.requestToken(ctx, v.DeviceCode, d.ClientID)
 		if accessToken == "" && err == nil {
 			if d.Clock.Now().After(v.ExpiresAt) {
 				err = errors.New("authentication timed out")
@@ -189,7 +186,7 @@ func (d *DeviceAuthenticator) GetAccessTokenForDevice(ctx context.Context, v *De
 
 		break
 	}
-	return accessToken, err
+	return accessToken, refreshToken, err
 }
 
 // OAuthTokenResponse contains the information returned after fetching an access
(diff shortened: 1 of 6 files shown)


internal/auth/authenticator.go
```diff
@@ -172,13 +172,10 @@ func (d *DeviceAuthenticator) VerifyDevice(ctx context.Context) (*DeviceVerifica
 
 // GetAccessTokenForDevice uses the device verification response to fetch an
 // access token.
-func (d *DeviceAuthenticator) GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (string, error) {
-	var accessToken string
-	var err error
-
+func (d *DeviceAuthenticator) GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (accessToken string, refreshToken string, err error) {
```
mdlayher, line 175:
```suggestion
func (d *DeviceAuthenticator) GetAccessTokenForDevice(ctx context.Context, v *DeviceVerification) (accessToken, refreshToken string, err error) {
```

internal/auth/authenticator.go
```diff
@@ -201,38 +198,67 @@ type OAuthTokenResponse struct {
...
 	err = json.NewDecoder(res.Body).Decode(tokenRes)
 	if err != nil {
-		return "", errors.Wrap(err, "error decoding token response")
+		return "", "", errors.Wrap(err, "error decoding token response")
+	}
+
+	return tokenRes.AccessToken, tokenRes.RefreshToken, nil
+}
+
+// RefreshToken for an access token
```
mdlayher, line 235: Note the meaning of each of the returned strings in the doc comment. Named return values are also doable but I tend to prefer better doc comments instead.

internal/auth/authenticator.go
```diff
@@ -201,38 +198,67 @@ type OAuthTokenResponse struct {
...
+func (d *DeviceAuthenticator) RefreshToken(ctx context.Context, token string) (string, string, error) {
```
mdlayher, line 236: It seems like this could effectively share the same logic with the initial GetToken calls, but with slightly differing form-encoded payload values.

internal/cmd/auth/login.go
```diff
@@ -70,6 +70,11 @@ func LoginCmd(ch *cmdutil.Helper) *cobra.Command {
 				return errors.Wrap(err, "error logging in")
 			}
 
+			err = writeRefreshToken(ctx, refreshToken)
+			if err != nil {
+				return errors.Wrap(err, "error logging in")
```
mdlayher, line 75:
```suggestion
				return errors.Wrap(err, "error refreshing token")
```

mdlayher: Perhaps the error on line 70 would need to be updated too?

internal/cmd/auth/login.go
```diff
@@ -150,3 +155,33 @@ func writeAccessToken(ctx context.Context, accessToken string) error {
...
 }
+
+func writeRefreshToken(ctx context.Context, refreshToken string) error {
+	configDir, err := config.ConfigDir()
+	if err != nil {
+		return err
+	}
+
+	_, err = os.Stat(configDir)
+	if os.IsNotExist(err) {
```
mdlayher, line 166:
```suggestion
	if errors.Is(err, os.ErrNotExist) {
```

internal/cmd/auth/login.go
```diff
@@ -150,3 +155,33 @@ func writeAccessToken(ctx context.Context, accessToken string) error {
...
+		err := os.MkdirAll(configDir, 0771)
```
mdlayher, line 167:
```suggestion
		err := os.MkdirAll(configDir, 0o771)
```

internal/cmd/auth/login.go
```diff
@@ -150,3 +155,33 @@ func writeAccessToken(ctx context.Context, accessToken string) error {
...
+		return err
+	}
+
+	tokenPath, err := config.RefreshTokenPath()
+	if err != nil {
+		return err
+	}
+
+	tokenBytes := []byte(refreshToken)
+	err = ioutil.WriteFile(tokenPath, tokenBytes, config.TokenFileMode)
```
mdlayher, line 181:
```suggestion
	err = os.WriteFile(tokenPath, tokenBytes, config.TokenFileMode)
```

internal/cmd/auth/logout.go
```diff
@@ -77,6 +77,18 @@ func deleteAccessToken() error {
 		}
 	}
 
+	refreshTokenPath, err := config.RefreshTokenPath()
+	if err != nil {
+		return err
+	}
+
+	err = os.Remove(refreshTokenPath)
+	if err != nil {
```
mdlayher, line 86:
```suggestion
	if err != nil && !errors.Is(err, os.ErrNotExist) {
```

internal/cmd/auth/refresh.go
```diff
@@ -0,0 +1,64 @@
...
+	var clientSecret string
+	var authURL string
+
+	cmd := &cobra.Command{
+		Use:   "refresh",
+		Args:  cobra.ExactArgs(0),
+		Short: "Refresh authentication with the PlanetScale API",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !printer.IsTTY {
+				return errors.New("The 'login' command requires an interactive shell")
```
mdlayher, line 23:
```suggestion
				return errors.New("The 'refresh' command requires an interactive shell")
```

internal/cmd/auth/refresh.go
```diff
@@ -0,0 +1,64 @@
...
+			// we explicitly stop here so we can replace the spinner with our success
+			// message.
+			end()
+			ch.Printer.Println("Successfully refreshed authentication.")
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&clientID, "client-id", auth.OAuthClientID, "The client ID for the PlanetScale CLI application.")
```
mdlayher, line 59: If this code is internal only to the pscale CLI anyway, what's the purpose of having flags for different oauth client ID/secret? Do we swap these for some kind of integration test?

internal/config/config.go
```diff
@@ -61,9 +62,33 @@ func New() (*Config, error) {
...
 
+	refreshTokenPath, err := RefreshTokenPath()
+	if err != nil {
+		return nil, err
+	}
+
+	stat, err = os.Stat(refreshTokenPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			log.Fatal(err)
```
mdlayher, line 73:
```suggestion
			return nil, err
```

internal/config/config.go
```diff
@@ -61,9 +62,33 @@ func New() (*Config, error) {
...
+		}
+	} else {
+		if stat.Mode()&^TokenFileMode != 0 {
+			err = os.Chmod(refreshTokenPath, TokenFileMode)
+			if err != nil {
+				log.Printf("Unable to change %v file mode to 0%o: %v", tokenPath, TokenFileMode, err)
+			}
+		}
+		refreshToken, err = ioutil.ReadFile(refreshTokenPath)
```
mdlayher, line 82:
```suggestion
		refreshToken, err = os.ReadFile(refreshTokenPath)
```

internal/config/config.go
```diff
@@ -61,9 +62,33 @@ func New() (*Config, error) {
...
+		if err != nil {
+			log.Fatal(err)
```
mdlayher, line 84:
```suggestion
			return nil, err
```

internal/config/config.go
```diff
@@ -106,6 +131,16 @@ func AccessTokenPath() (string, error) {
...
 }
 
+// RefreshToken is the path for the refresh token file
+func RefreshTokenPath() (string, error) {
+	dir, err := ConfigDir()
+	if err != nil {
+		return "", err
+	}
+
+	return path.Join(dir, "refresh-token"), nil
```
mdlayher, line 141:
```suggestion
	return filepath.Join(dir, "refresh-token"), nil
```

Or else this won't work on Windows!

mdlayher (commented):
I like the idea, mostly some Go nitpicks and such but a few questions too.

PR #447, opened 2022-01-05
Switch service token name to be service token ID

iheanyi:
Before, we were showing the ID of a service token as a name, when in reality, we should just keep it as the `ID` in the output. This makes a couple of changes. First, it changes the header for a service token to be `ID` instead of `Name`. Secondly, it changes the flag to be `--service-token-id` instead of `--service-token-name`, with the last one being deprecated and hidden. Users will get warned when using the deprecated flag to use the new format, so this gives our users ample time to switch over.


no-itsbackpack (approved):
✅

PR #448, opened 2022-01-12
Pin to 3.4.0 for licensed.

iheanyi:
3.4.1 seems to be causing issues, so let's see if using 3.4.0 fixes us up.


nickvanw:
I think we can close this with #451 merged and working 😁

PR #465, opened 2022-02-24
Add `role` parameter to shell & connect commands

Phanatic:
This PR adds a `--role` parameter as a hidden flag to the `pscale shell` and `pscale connect` commands. 

We can now prevent accidental inserts & deletes from a database when using shell for read only queries.

``` bash
 go run cmd/pscale/main.go shell  withacls main --org phani --role reader
withacls/main> show tables;
+--------------------+
| Tables_in_withacls |
+--------------------+
| Customers          |
+--------------------+
withacls/main> select count(*) from Customers;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
withacls/main> Insert into Customers(ID) values (12);
ERROR 1045 (28000): target: withacls.-.primary: vttablet: rpc error: code = PermissionDenied desc = Insert command denied to user 'planetscale-reader' for table 'Customers' (ACL check error) (CallerID: planetscale-reader)

withacls/main> Delete from Customers;
ERROR 1045 (28000): target: withacls.-.primary: vttablet: rpc error: code = PermissionDenied desc = DeleteLimit command denied to user 'planetscale-reader' for table 'Customers' (ACL check error) (CallerID: planetscale-reader)
withacls/main>
```

### DIFF
--- a/internal/cmd/shell/shell.go
+++ b/internal/cmd/shell/shell.go
@@ -83,6 +83,14 @@ second argument:
 				}
 			}
 
+			role := cmdutil.Reader
+			if flags.role != "" {
+				role, err = cmdutil.RoleFromString(flags.role)
+				if err != nil {
+					return err
+				}
+			}
+
 			// check whether database and branch exist
 			_, err = client.DatabaseBranches.Get(ctx, &ps.GetDatabaseBranchRequest{
 				Organization: ch.Config.Organization,
@@ -106,7 +114,7 @@ second argument:
 			}
 
 			proxyOpts := proxy.Options{
-				CertSource: proxyutil.NewRemoteCertSource(client, flags.role),
+				CertSource: proxyutil.NewRemoteCertSource(client, role),
 				LocalAddr:  localAddr,
 				RemoteAddr: flags.remoteAddr,
 				Instance:   fmt.Sprintf("%s/%s/%s", ch.Config.Organization, database, branch),
--- a/internal/cmd/shell/shell.go
+++ b/internal/cmd/shell/shell.go
@@ -197,7 +197,7 @@ second argument:
 	cmd.PersistentFlags().StringVar(&flags.remoteAddr, "remote-addr", "",
 		"PlanetScale Database remote network address. By default the remote address is populated automatically from the PlanetScale API.")
 	cmd.PersistentFlags().StringVar(&flags.role, "role",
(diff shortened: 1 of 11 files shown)


internal/proxyutil/remotesource.go
```diff
@@ -6,6 +6,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"fmt"
+	"github.com/planetscale/cli/internal/cmdutil"
 	"time"
```
fatih, line 10: Could you move the import path next to the third party ones?

internal/cmd/shell/shell.go
```diff
@@ -188,10 +196,10 @@ second argument:
 		"", "Local address to bind and listen for connections. By default the proxy binds to 127.0.0.1 with a random port.")
 	cmd.PersistentFlags().StringVar(&flags.remoteAddr, "remote-addr", "",
 		"PlanetScale Database remote network address. By default the remote address is populated automatically from the PlanetScale API.")
-	cmd.PersistentFlags().BoolVar(&flags.readOnly, "read-only",
-		false, "Connect to database in read only mode.")
+	cmd.PersistentFlags().StringVar(&flags.role, "role",
+		"reader", "allowed values are : reader, writer, readwriter, admin, default is reader")
```
fatih, line 200: Would like to see a text message that describes the flag. Such as: 

`Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is reader.`

internal/cmdutil/cmdutil.go
```diff
@@ -22,6 +22,45 @@ import (
 const WarnAuthMessage = "not authenticated yet. Please run 'pscale auth login' " +
 	"or create a service token with 'pscale service-token create'"
 
+type PasswordRole int
+
+const (
+	Reader PasswordRole = iota
+	Writer
+	ReadWriter
+	Administrator
```
fatih, line 31: When we import these constants, they don't tell exactly what they are. especially if its `cmdutil.Reader` or `cmdutil.Writer`. They look similar to `io.Reader` or `io.Writer`, though it's something entirely different. 

What do you think of renaming these by adding a `Role` suffix? : 

```suggestion
	ReaderRole PasswordRole = iota
	WriterRole
	ReadWriterRole
	AdministratorRole
```

That way it would read as `cmdutil.ReaderRole` or `cmdutil.WriterRole`

fatih (approved):
This looks great 👍🏼

PR #471, opened 2022-03-18
Switch to using a keyring for password storage if available

dbussink:
This logic switches to using a keyring like for example keychain on MacOS if that is available on the platform.

This provides better and more secure secret storage for the token that gets generated for a user.

If no keyring implementation is available, things keep working like they work today.


internal/config/config.go
```diff
@@ -1,14 +1,15 @@
 package config
 
 import (
-	"errors"
```
mdlayher, line 4: For open source I'd encourage us to stick with what the broader community is already using, and these days that'd be the stdlib package. I'm not sure stack traces would be super helpful in a CLI app.

internal/config/config.go
```diff
@@ -130,3 +104,179 @@ func RootGitRepoDir() (string, error) {
...
+		return string(item.Data), nil
+	}
+
+	if errors.Is(err, keyring.ErrKeyNotFound) {
+		// Migrate to keychain
+		accessToken, tokenErr := readAccessTokenPath()
+		if len(accessToken) > 0 && tokenErr == nil {
+			return migrateAccessToken(ring, accessToken)
+		}
+		return "", nil
```
mdlayher, line 127: I feel like this should return an error of some sort, yeah?

internal/config/config.go
```diff
@@ -130,3 +104,179 @@ func RootGitRepoDir() (string, error) {
...
+		Data: accessToken,
+	})
+	if err != nil {
+		return "", err
+	}
+	path, err := accessTokenPath()
+	if err != nil {
+		return "", err
+	}
+	err = os.Remove(path)
```
mdlayher, line 145: I'm not sure how I feel about automatically removing a file on behalf of users. Should the migration be opt-in? Do we consider this file to be an implementation detail of the CLI that users should never interact with?

internal/config/config.go
```diff
@@ -130,3 +104,179 @@ func RootGitRepoDir() (string, error) {
...
+	})
+}
+
+func accessTokenPath() (string, error) {
+	dir, err := ConfigDir()
+	if err != nil {
+		return "", err
+	}
+
+	return path.Join(dir, keyringKey), nil
```
mdlayher, line 195:
```suggestion
	return filepath.Join(dir, keyringKey), nil
```

filepath properly handles Windows too.

mdlayher (commented):
LGTM overall, I see there's a mix of old and new code so consider some comments nonblocking.

internal/config/config.go
dbussink: @mdlayher I thought that originally as well, but that breaks. The reason is that we create a config object always on the CLI startup, also if there's no authentication token configured yet. So right now today an empty token means no token and that case should not error since it breaks things.

It might be good to make that more explicit, but that's I think something for a separate change.

internal/config/config.go
dbussink: > Do we consider this file to be an implementation detail of the CLI that users should never interact with?

Yes, the file is an implementation detail and I think we should do this for the user so they immediately benefit from a better security mechanism being used.

internal/config/config.go
dbussink: Ah, this was moved code but I think it's good to fix it up then.

internal/config/config.go
```diff
@@ -130,3 +104,179 @@ func RootGitRepoDir() (string, error) {
...
+	} else {
+		if stat.Mode()&^tokenFileMode != 0 {
+			err = os.Chmod(tokenPath, tokenFileMode)
+			if err != nil {
+				log.Printf("Unable to change %v file mode to 0%o: %v", tokenPath, tokenFileMode, err)
+			}
+		}
+		accessToken, err = ioutil.ReadFile(tokenPath)
+		if err != nil {
+			log.Fatal(err)
```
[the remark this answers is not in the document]
dbussink: I think this is similar that it might be dependent on this working / treating a missing file as "no token configured".

That all could be improved, but I think that's something separate.

internal/config/config.go
dbussink: @mdlayher We're already using error wrapping in a bunch of other places and would prefer to keep that consistent then. I think the value for us more in the wrapping than in stack traces.

mdlayher: Oh I see now, this is already in use in other files in the repo. SGTM.

internal/config/config.go
mdlayher: Sure, mind just dropping a one-liner TODO comment to note this behavior? An empty, empty return usually looks like a bug.

internal/config/config.go
mdlayher: I'll defer to @fatih and @iheanyi too since I think they've both got a lot more time in this than I do.

mdlayher (approved):
LGTM overall.

internal/config/config.go
fatih: I'm worried this might break CI users. Sometimes automations happen by mounting a secret file or something alike, so if the path is in such a path and the CI expects it somehow, I'm not sure how it would behave. Another edge case, sometimes the disk for certain files are read-only, so this might fail as well? 

I wonder if we should add a prompt or a flag that removes it from their disk. Something like: "Would you like to remove your old token?" or a `--force-remove-auth` flag.

fatih: Another feedback: I think we should not error here if we can't remove it. Because we already successfully obtained the secret. I think printing a warning such as `"Couldn't remove old access token due to error: ..."`

fatih (approved):
LGTM with a comment on the migration path. I think we shouldn't remove it forcefully.

internal/config/config.go
dbussink: @fatih I don't see how a CI system would have something that here is picked up as a working keyring? It would only do that if there's a working keyring implementation which normally is not the case.

Even though if there is a keyring, in general CI systems would treat disks as ephemeral? Like it would drop the secret into place, but if there's a working keyring it could move it there but it still works.

Then on a next CI run, the file would be put back in place again and things keep working?

The reason I'm pretty strong on removing it, is because without it the vast majority of cases would never remove the file then. It would be lingering there forever. 

> I think printing a warning such as `"Couldn't remove old access token due to error: ..."`

I think similarly here, I worry that people ignore a warning and that we still have a valid token on disk lingering around that still works.

fatih: > The reason I'm pretty strong on removing it, is because without it the vast majority of cases would never remove the file then. It would be lingering there forever.

That's a good point. 

> Even though if there is a keyring, in general CI systems would treat disks as ephemeral? Like it would drop the secret into place, but if there's a working keyring it could move it there but it still works.

They would treat as ephemeral, thought some CI systems also just attach/mount an EBS for the underlying disk. It can go crazy. For GitHub actions or similar CI's you're right, but people with a more traditional setup, such as using Jenkins might not have the same environment.

> I think similarly here, I worry that people ignore a warning and that we still have a valid token on disk lingering around that still works.

That's a good call, but I think it might still break people's workflows. What if we do this: Let's not return an error, print a warning. Deploy `pscale` and remove the warning in like 1/2 weeks later. That would show us if we missed any edge cases.

dbussink: > They would treat as ephemeral, thought some CI systems also just attach/mount an EBS for the underlying disk. It can go crazy. For GitHub actions or similar CI's you're right, but people with a more traditional setup, such as using Jenkins might not have the same environment.

Yeah, but the same reasoning here applies? If the machine is a long term stateful thing and it has a keyring, it means that the keyring also would keep working?

> That's a good call, but I think it might still break people's workflows. What if we do this: Let's not return an error, print a warning. Deploy pscale and remove the warning in like 1/2 weeks later. That would show us if we missed any edge cases.

What about a more general changed approach here that does include removing the file (and erroring it fails), but we don't do that right now.

We first release the CLI now without immediately deleting it, to see if that causes any adverse effects for people. We then release a future new version in a few weeks that deletes the token from disk if it successfully reads from the keyring? 

I think that helps address all situations then?

fatih: > Yeah, but the same reasoning here applies? If the machine is a long term stateful thing and it has a keyring, it means that the keyring also would keep working?

True, but there are also systems like in K8S where you can attach/mount stateful disks, even though the underlying compute instance might be not stateful (i.e: Pod). I don't know how various Keyring environments work of course, but I think being careful here wouldn't hurt us. 

> We first release the CLI now without immediately deleting it, to see if that causes any adverse effects for people. We then release a future new version in a few weeks that deletes the token from disk if it successfully reads from the keyring?
> I think that helps address all situations then?

I like it a lot, let's do it 👍🏼

PR #483, opened 2022-03-31
Error when half of a service token is passed

nickvanw:
In order to use authentication with a Service Token, you have to pass both the Token ID and the Token. If you also have a local session as a user, it will fall back to using that if you pass one of the two.

Instead, we should error if you've passed one or the other, which is significantly less confusing.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -78,11 +78,11 @@ func (c *Config) NewClientFromConfig(clientOpts ...ps.ClientOption) (*ps.Client,
 	}
 
 	if c.ServiceToken == "" && c.ServiceTokenID != "" {
-		return nil, errors.New("service token ID was passed, but service token was not")
+		return nil, errors.New("both --service-token and --service-token-id are required for service token authentication")
 	}
 
 	if c.ServiceToken != "" && c.ServiceTokenID == "" {
-		return nil, errors.New("service token was passed, but service token ID was not")
+		return nil, errors.New("both --service-token and --service-token-id are required for service token authentication")
 	}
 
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -77,6 +77,14 @@ func (c *Config) NewClientFromConfig(clientOpts ...ps.ClientOption) (*ps.Client,
 		ps.WithBaseURL(c.BaseURL),
 	}
 
+	if c.ServiceToken == "" && c.ServiceTokenID != "" {
+		return nil, errors.New("service token ID was passed, but service token was not")
```
mscoutermarsh, line 81:
```suggestion
		return nil, errors.New("both --service-token and --service-token-id are required for service token authentication")
```

internal/config/config.go
```diff
@@ -77,6 +77,14 @@ func (c *Config) NewClientFromConfig(clientOpts ...ps.ClientOption) (*ps.Client,
...
+	}
+
+	if c.ServiceToken != "" && c.ServiceTokenID == "" {
+		return nil, errors.New("service token was passed, but service token ID was not")
```
mscoutermarsh, line 85:
```suggestion
		return nil, errors.New("both --service-token and --service-token-id are required for service token authentication")
```

mscoutermarsh (approved):

nickvanw:
```./pscale --service-token blah org list
!! WARNING: You are using a self-compiled binary which is not officially supported.
!! To dismiss this warning, set PSCALE_DISABLE_DEV_WARNING=true

Error: both --service-token and --service-token-id are required for service token authentication
```

PR #484, opened 2022-04-04
Add ARMv7 Linux builds

jgreet:
A user asked for ARMv7 Linux builds, this config adds only that build,

Not sure if there's a better way to check this but currently running `goreleaser build` on main locally builds these targets:

```sh
 • building binaries
      • building                  binary=dist/pscale_darwin_amd64/pscale
      • building                  binary=dist/pscale_darwin_arm64/pscale
      • building                  binary=dist/pscale_linux_386/pscale
      • building                  binary=dist/pscale_linux_amd64/pscale
      • building                  binary=dist/pscale_linux_arm64/pscale
      • building                  binary=dist/pscale_windows_386/pscale.exe
      • building                  binary=dist/pscale_windows_amd64/pscale.exe
      • building                  binary=dist/pscale_windows_arm64/pscale.exe
```

And for this branch, we add exactly one more:
```sh
      • building                  binary=dist/pscale_darwin_amd64/pscale
      • building                  binary=dist/pscale_darwin_arm64/pscale
      • building                  binary=dist/pscale_linux_386/pscale
      • building                  binary=dist/pscale_linux_amd64/pscale
      • building                  binary=dist/pscale_linux_arm64/pscale
+     • building                  binary=dist/pscale_linux_arm_7/pscale
      • building                  binary=dist/pscale_windows_386/pscale.exe
      • building                  binary=dist/pscale_windows_amd64/pscale.exe
      • building                  binary=dist/pscale_windows_arm64/pscale.exe
```

Followed https://goreleaser.com/customization/build/

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -12,6 +12,17 @@ builds:
       - linux
       - windows
       - darwin
+    goarch:
+      - 386
+      - amd64
+      - arm
+      - arm64
+    goarm:
+      - 7
+    ignore:
+      - goos: windows
+        goarch: arm
+        goarm: 7
     main: ./cmd/pscale/main.go
     ldflags:
      - -s -w -X main.version={{.Version}} -X main.commit={{.ShortCommit}} -X main.date={{.Date}}


.goreleaser.yml
```diff
@@ -12,6 +12,17 @@ builds:
...
+      - 386
+      - amd64
+      - arm
+      - arm64
+    goarm:
+      - 7
+    ignore:
+      - goos: windows
+        goarch: arm
+        goarm: 7
```
fatih, line 25: Why do we ignore this? Could you add a comment maybe to the yaml file?

fatih (approved):
Looks good to me 👍🏼 I added a comment for the ignore directive. It would be nice to have a comment on why we're ignoring it.

.goreleaser.yml
```diff
@@ -12,6 +12,17 @@ builds:
       - linux
       - windows
       - darwin
+    goarch:
+      - 386
+      - amd64
+      - arm
+      - arm64
+    goarm:
+      - 7
```
dbussink, line 21: I think it's better to drop this bit and use the default ARMv6, see also https://go.dev/doc/install/source#environment.

That will also work on ARMv7 and makes it work a bit more generally also for older ARM hardware. We're not trying to eek out the last bit of performance here then.

dbussink:
```
GOARM=6: use VFPv1 only; default if cross compiling; usually ARM11 or better cores (VFPv2 or better is also supported)
```

.goreleaser.yml
jgreet: I didn't think there were any current 32bit ARM Windows versions left but apparently [Embedded Compact and IoT Core](https://en.wikipedia.org/wiki/ARM_architecture_family#32-bit_operating_systems) are still supported.

I'll take this out.

jgreet:
This now adds uses the defaults for 32bit ARM and adds 2 builds, 1 each for Linux and Windows

```
   • building binaries
      • building                  binary=dist/pscale_darwin_amd64/pscale
      • building                  binary=dist/pscale_darwin_arm64/pscale
      • building                  binary=dist/pscale_linux_386/pscale
      • building                  binary=dist/pscale_linux_amd64/pscale
      • building                  binary=dist/pscale_linux_arm64/pscale
+     • building                  binary=dist/pscale_linux_arm_6/pscale
      • building                  binary=dist/pscale_windows_386/pscale.exe
      • building                  binary=dist/pscale_windows_amd64/pscale.exe
      • building                  binary=dist/pscale_windows_arm64/pscale.exe
+     • building                  binary=dist/pscale_windows_arm_6/pscale.exe
```

PR #489, opened 2022-04-07
internal/auth: construct form-encoded bodies with url.Values

mdlayher:
Signed-off-by: Matt Layher <mdlayher@planetscale.com>

While looking at this on stream, we discovered that the existing test fixtures appeared to be incorrect. The server does accept properly URL-encoded POST form bodies as are now provided when we ultimately encode the `url.Values` maps.

I've built a local copy of `pscale` and can confirm that `pscale auth login` continues to work with the new request format.

I do think we should follow up on the auth server side to check out what sort of logic is being used to decode these requests, so we can verify that all possible inputs are covered by tests.

/cc @iheanyi @fatih

### DIFF
--- a/internal/auth/authenticator.go
+++ b/internal/auth/authenticator.go
@@ -3,8 +3,6 @@ package auth
 import (
 	"context"
 	"encoding/json"
-	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/auth/authenticator.go
```diff
@@ -129,9 +127,14 @@ func New(client *http.Client, clientID, clientSecret string, opts ...Authenticat
 
 // VerifyDevice performs the device verification API calls.
 func (d *DeviceAuthenticator) VerifyDevice(ctx context.Context) (*DeviceVerification, error) {
-	oauthScopes := []string{"read_databases", "write_databases", "read_user", "read_organization"}
-	payload := strings.NewReader(fmt.Sprintf("client_id=%s&scope=%s", d.ClientID, url.PathEscape(strings.Join(oauthScopes, " "))))
-	req, err := d.NewFormRequest(ctx, http.MethodPost, "oauth/authorize_device", payload)
+	req, err := d.newFormRequest(ctx, "oauth/authorize_device", url.Values{
+		"client_id": []string{d.ClientID},
+		// Scopes are concatenated with strings and ultimately URL-encoded for
+		// use by the server.
```
dbussink, line 133: https://datatracker.ietf.org/doc/html/rfc6749#section-3.3 for if you want to throw in an RFC definition :wink:.

dbussink, line 133: https://datatracker.ietf.org/doc/html/rfc6749#section-3.3 for if you want to throw in an RFC definition :wink:.

dbussink (approved):

PR #498, opened 2022-05-10
goreleaser: stop disabling CGO

nickvanw:
Testing this out to see if everything builds properly even with cross-compilation.


iheanyi (approved):
🚀

dbussink:
https://github.com/planetscale/cli/pull/523 is the alternative approach for this which I think would work.

PR #523, opened 2022-06-29
Release MacOS with CGO enabled

dbussink:
This updates the `goreleaser` usage to use https://github.com/goreleaser/goreleaser-cross which provides a cross compilation toolchain so we can enable CGO for the MacOS binary.

https://github.com/planetscale/discussion/discussions/181 talks more about the issue this helps with which was also reported upstream in https://github.com/golang/go/issues/52839 and I suspect https://github.com/golang/go/issues/43398 is also very related.

We can work around this by building with `CGO_ENABLED=1` but we do need a cross compilation setup which https://github.com/goreleaser/goreleaser-cross provides.


Makefile
```diff
@@ -11,6 +11,7 @@ endif
 REPO=planetscale
 NAME=pscale
 BUILD_PKG=github.com/planetscale/cli/cmd/pscale
+GOLANG_CROSS_VERSION  ?= v1.18.3
```
fatih, line 14: I think this is the version of goreleaser-cross. It looks like they try to follow Go's versioning, but changing this also changes the goreleaser version. Previously we could use a specific version, I believe we can't do that anymore.

I say let's rename this to `GORELEASE_CROSS_VERSION` to indicate that this is the version of the docker image.

fatih (approved):

Makefile
dbussink: @fatih Updated!

PR #528, opened 2022-06-29
Add pagination and limits for audit logs in the CLI.

iheanyi:
This pull request adds support for paginating audit logs via the CLI using our new cursor-based pagination approach. It adds two new flags, `--starting-after` and `--limit` so users can control the output to their liking. 👀 

cc @mscoutermarsh

Depends on https://github.com/planetscale/planetscale-go/pull/131 landing first.


fatih (approved):
Looks good to me 👍🏼

PR #571, opened 2022-10-17
fix(ci): update docker creds for latest goreleaser-cross

joemiller:
goreleaser-cross now requires docker creds provided in a JSON file.

We could also add a new CI job that runs on non-main branches that execs goreleaser with the `--snapshot` flag. That will run goreleaser but not publish a release. I've found this helpful on other projects since goreleaser has a tendency to change config options frequently. I added the `GORELEASER_EXTRA_ARGS` to help facilitate this with local testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ build:
 	@go build -trimpath ./...
 
 .PHONY: lint
-lint: 
+lint:
 	@go install honnef.co/go/tools/cmd/staticcheck@HEAD
 	@staticcheck ./...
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -64,8 +64,10 @@ release:
 		-e DOCKER_USERNAME=${DOCKER_USERNAME} \
 		-e DOCKER_PASSWORD=${DOCKER_PASSWORD} \
```
fatih, line 65: Should we remove these two lines?

iheanyi (approved):
Let's see what happens.

PR #582, opened 2022-11-16
dump/restore: set number of threads to be configurable

nickvanw:
This allows a user to pick the number of threads to use when dumping and restoring. Defaults are unchanged for dump, but are set low (1) for restoring, to ensure that we don't overwhelm a small database. Users can configure that to be higher if they happen to want higher concurrency.

```
$ pscale db restore-dump --help

Restore your database from a local dump directory

Usage:
  pscale database restore-dump <database> <branch> [options] [flags]

Flags:
      --dir string          Directory containing the files to be used for the restore (required)
  -h, --help                help for restore-dump
      --local-addr string   Local address to bind and listen for connections. By default the proxy binds to 127.0.0.1 with a random port.
      --overwrite-tables    If true, will attempt to DROP TABLE before restoring.
      --threads int         Number of concurrent threads to use to restore the database. (default 1)
```

### DIFF
--- a/internal/cmd/database/dump.go
+++ b/internal/cmd/database/dump.go
@@ -48,7 +48,7 @@ func DumpCmd(ch *cmdutil.Helper) *cobra.Command {
 		"Comma separated string of WHERE clauses to filter the tables to dump. Only used when you specify tables to dump. Default is not to filter dumped tables.")
 	cmd.PersistentFlags().StringVar(&f.output, "output", "",
 		"Output directory of the dump. By default the dump is saved to a folder in the current directory.")
-	cmd.PersistentFlags().IntVar(&f.threads, "threads", 16, "Number of concurrent threads to use to restore the database.")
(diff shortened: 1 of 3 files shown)


internal/cmd/database/dump.go
```diff
@@ -47,6 +48,7 @@ func DumpCmd(ch *cmdutil.Helper) *cobra.Command {
 		"Comma separated string of WHERE clauses to filter the tables to dump. Only used when you specify tables to dump. Default is not to filter dumped tables.")
 	cmd.PersistentFlags().StringVar(&f.output, "output", "",
 		"Output directory of the dump. By default the dump is saved to a folder in the current directory.")
+	cmd.PersistentFlags().IntVar(&f.threads, "threads", 16, "Number of concurrent threads to use to restore the database.")
```
orware, line 51: Not a showstopper, but I think the flag description may need to read "to dump the database" there at the end in the dump.go file.

PR #585, opened 2022-11-22
Restore-dump: surface errors

nickvanw:
When we hit an error inserting data, the current behavior of `restore-dump` is to log it, but not return it - this means that the command will exit with a 0 error code even if there was an error in the restore path, which can be misleading.

This will now emit an error per-table, and surfaces the first one that happens all the way up to the top of the command. 

cc @fatih. Not wedded to the approach, but it's straightforward.

### DIFF
--- a/internal/dumper/loader.go
+++ b/internal/dumper/loader.go
@@ -74,6 +74,8 @@ func (l *Loader) Run(ctx context.Context) error {
 		files.tables[i], files.tables[j] = files.tables[j], files.tables[i]
 	}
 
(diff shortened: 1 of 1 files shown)


iheanyi (approved):

fatih:
LGTM 👍🏼  I think it's time we switch to https://pkg.go.dev/golang.org/x/sync/errgroup to collect the errors.

fatih:
Opened a PR to switch to errgroup: https://github.com/planetscale/cli/pull/586

PR #597, opened 2022-12-19
Ensure we create the config directory before we write to it

### LINKED ISSUE #596: `pscale auth login` fails on a new machine
I'm setting up on my laptop. I installed `pscale` via `brew`. Then, running `pscale auth login` does the oauth dance with the browser fine, but then fails with the error: 

```
~ pscale auth login

Confirmation Code:<code>

If something goes wrong, copy and paste this URL into your browser: https://auth.planetscale.com/oauth/device?user_code=<coid>

Successfully logged in.
Error: open /Users/threepointone/.config/planetscale/pscale.yml: no such file or directory
```

This happened multiple times. I fixed it by creating the `/Users/threepointone/.config/planetscale` folder, so I suspect the bug is that the cli is trying to create the `pscale.yml` file without first checking whether the directory exists.

fooforge:
WIP

Fixes #596

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -290,22 +290,7 @@ func deleteAccessTokenPath() error {
 	return nil
 }
 
-func writeAccessTokenPath(accessToken string) error {
-	configDir, err := ConfigDir()
-	if err != nil {
-		return err
-	}
-
-	_, err = os.Stat(configDir)
-	if os.IsNotExist(err) {
-		err := os.MkdirAll(configDir, 0771)
-		if err != nil {
-			return errors.Wrap(err, "error creating config directory")
-		}
-	} else if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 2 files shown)


internal/cmd/auth/login.go
```diff
@@ -99,6 +99,12 @@ func writeDefaultOrganization(ctx context.Context, accessToken, authURL string)
 		return err
 	}
 
+	// If the configuration directory doesn't exist, we create it.
+	err = config.CreateConfigDir()
+	if err != nil {
+		return err
+	}
```
dbussink, line 106: What about moving this into `WriteDefault()` instead? Where we do the actual file write?

fooforge: Good call, thanks!

dbussink (approved):

PR #641, opened 2023-03-24
Add `syft` to goreleaser container

nickvanw:
This is so that we can produce SBOMs as part of the release process.


iheanyi (approved):
Good looks!

PR #658, opened 2023-04-26
Add local development instructions

mscoutermarsh:
Giving an example on how to build & run locally.


iheanyi (approved):
They also can use `go run cmd/pscale/main.go <command>` if they don't feel like building the binary.

Issue #556, opened 2022-09-07
Better error for db list with a new service token


jgreet:
From https://github.com/planetscale/discussion/discussions/267

Authenticating `pscale database list` with a service token that lacks any accesses shows an error message that can be confusing if you have an existing database but don't know about access grants yet.

```
$ pscale service-token create
$ pscale db list --service-token pscale_tkn_xxx --service-token-id 6hnhqpmd0fv7
No databases have been created yet.
```

PR #659, opened 2023-04-26
Improve error message for `db list` when org missing or not authed

mscoutermarsh:
Fixes: https://github.com/planetscale/cli/issues/556

Improves the error message for cases where the user doesn't have permission to access it.

We can't get more specific than this since we do not give more detailed info via the API.


iheanyi (approved):
💯

Issue #652, opened 2023-04-19
database dump doesn't dump invisible columns


wooque:
Ran `pscale database dump test main` but data sql (in my case `test.events.00001.sql`) doesn't include invisible column (in my case `event_data` )

mscoutermarsh:
Thanks for reporting! Fix was just merged in, will be included in the next release

Issue #645, opened 2023-03-31
[Bug] pscale database dump / restore-dump commands do not correctly ignore GENERATED STORED columns for INSERTs


orware:
We had a ticket arrive today that provided an example that demonstrated that our current `pscale database dump` / `pscale database restore-dump` commands aren't correctly ignoring `GENERATED` columns that are also `STORED`.

To reproduce this issue, within a test database you can use the following simplified example table definition:

```sql
CREATE TABLE `stock_trades` (
  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  `unit_price` decimal(13,4) DEFAULT NULL,
  `quantity` decimal(13,4) DEFAULT NULL,
  `total_value` decimal(15,4) GENERATED ALWAYS AS ((`quantity` * `unit_price`)) STORED,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

And example rows:

```sql
INSERT INTO `stock_trades`(`id`,`unit_price`,`quantity`) VALUES
(1,12.1000,2.1000),
(2,74.3200,99.0000),
(3,56.4000,34.0000),
(4,230.3100,22.0000),
(5,133.3300,3.0000),
(6,0.0100,104.0000),
(7,11.2000,12.0000),
(8,131.4400,55.0000),
(9,18.8000,34.0000),
(10,99.2000,2.0000);
```

And you can observe that if you run a `SELECT * from stock_trades;` query from the web console that you can see the generated `total_value` column included.

If you run a `pscale database dump` command though, the generated schema file for the table above is ok.

However, the file that inserts the rows back into the table is not ok because it includes the stored `GENERATED` column as if it is a normal column:

```sql
INSERT INTO `stock_trades`(`id`,`unit_price`,`quantity`,`total_value`) VALUES
(1,12.1000,2.1000,25.4100),
(2,74.3200,99.0000,7357.6800),
(3,56.4000,34.0000,1917.6000),
(4,230.3100,22.0000,5066.8200),
(5,133.3300,3.0000,399.9900),
(6,0.0100,104.0000,1.0400),
(7,11.2000,12.0000,134.4000),
(8,131.4400,55.0000,7229.2000),
(9,18.8000,34.0000,639.2000),
(10,99.2000,2.0000,198.4000);
```

[...]

PR #666, opened 2023-04-28
Dump: support invisible columns

mscoutermarsh:
Fixes: https://github.com/planetscale/cli/issues/652
Fixes: https://github.com/planetscale/cli/issues/645

Originally, the dump command grabbed the column names by running `select *`.

This does not work with invisible columns, since `select *` will only return visible columns.

This PR updates `dump` to use `SHOW FIELDS` which returns invisible columns.

```
invisible/main> show fields from mike;
+-------+------+------+-----+---------+-----------+
| Field | Type | Null | Key | Default | Extra     |
+-------+------+------+-----+---------+-----------+
| id    | int  | NO   | PRI | NULL    |           |
| test  | int  | YES  |     | NULL    | INVISIBLE |
+-------+------+------+-----+---------+-----------+
```

We then use that list to generate the dump.

### DIFF
--- a/internal/dumper/dumper.go
+++ b/internal/dumper/dumper.go
@@ -408,9 +408,9 @@ func (d *Dumper) dumpableFieldNames(conn *Connection, table string) ([]string, e
 		name := t[0].String()
 		extra := t[5].String()
 
-		// Can be either "VIRTUAL GENERATED" or "VIRTUAL STORED"
+		// Can be either "VIRTUAL GENERATED" or "STORED GENERATED"
 		// https://dev.mysql.com/doc/refman/8.0/en/show-columns.html
-		if strings.Contains(extra, "VIRTUAL") {
+		if strings.Contains(extra, "GENERATED") {
 			// Skip generated columns
 			continue
 		} else {
--- a/internal/dumper/dumper.go
+++ b/internal/dumper/dumper.go
@@ -391,7 +391,7 @@ func (d *Dumper) filterDatabases(conn *Connection, filter *regexp.Regexp, invert
 	return databases, nil
 }
 
-// dumpableFieldNames returns a map that contains valid field names for the dump
+// dumpableFieldNames returns a slice that contains valid field names for the dump.
 func (d *Dumper) dumpableFieldNames(conn *Connection, table string) ([]string, error) {
 	qr, err := conn.Fetch(fmt.Sprintf("SHOW FIELDS FROM `%s`", table))
 	if err != nil {
--- a/internal/dumper/dumper.go
+++ b/internal/dumper/dumper.go
@@ -223,42 +223,27 @@ func (d *Dumper) dumpTable(conn *Connection, database string, table string) erro
 	var where string
 	var selfields []string
 
-	isGenerated, err := d.generatedFields(conn, table)
-	if err != nil {
-		return err
-	}
-
 	fields := make([]string, 0)
 	{
-		cursor, err := conn.StreamFetch(fmt.Sprintf("SELECT * FROM `%s`.`%s` LIMIT 1", database, table))
+		flds, err := d.dumpableFieldNames(conn, table)
 		if err != nil {
 			return err
 		}
 
-		flds := cursor.Fields()
 		for _, fld := range flds {
-			d.log.Debug("dump", zap.Any("filters", d.cfg.Filters), zap.String("table", table), zap.String("field_name", fld.Name))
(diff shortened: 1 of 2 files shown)


internal/dumper/dumper.go
```diff
@@ -223,43 +223,28 @@ func (d *Dumper) dumpTable(conn *Connection, database string, table string) erro
...
 
-	isGenerated, err := d.generatedFields(conn, table)
-	if err != nil {
-		return err
-	}
-
 	fields := make([]string, 0)
 	{
-		cursor, err := conn.StreamFetch(fmt.Sprintf("SELECT * FROM `%s`.`%s` LIMIT 1", database, table))
+		flds, err := d.dumpableFieldNames(conn, table)
```
mscoutermarsh, line 228: This is the primary change. This new function uses `show fields` to grab all the dumpable field names (any field that isn't generated).

internal/dumper/dumper_test.go
```diff
@@ -114,14 +125,13 @@ func TestDumper(t *testing.T) {
...
 		},
 		Rows: [][]sqltypes.Value{
-			{
-				sqltypes.MakeTrusted(querypb.Type_VARCHAR, []byte("not_deleted")),
-				sqltypes.MakeTrusted(querypb.Type_BLOB, []byte("int")),
-				sqltypes.MakeTrusted(querypb.Type_VARCHAR, []byte("YES")),
-				sqltypes.MakeTrusted(querypb.Type_BINARY, []byte("")),
-				sqltypes.MakeTrusted(querypb.Type_NULL_TYPE, []byte("NULL")),
-				sqltypes.MakeTrusted(querypb.Type_VARCHAR, []byte("VIRTUAL GENERATED")),
-			},
```
mscoutermarsh, line 124: Previously this stub's purpose was only for filtering out the single generated column. Now it's being used for grabbing all the column names.

internal/dumper/dumper.go
```diff
@@ -424,10 +408,13 @@ func (d *Dumper) generatedFields(conn *Connection, table string) (map[string]boo
 		name := t[0].String()
 		extra := t[5].String()
 
-		// Can be either "VIRTUAL GENERATED" or "VIRTUAL STORED"
+		// Can be either "VIRTUAL GENERATED" or "STORED GENERATED"
 		// https://dev.mysql.com/doc/refman/8.0/en/show-columns.html
-		if strings.Contains(extra, "VIRTUAL") {
-			fields[name] = true
+		if strings.Contains(extra, "GENERATED") {
```
mscoutermarsh, line 413: I believe this has been a bug for a while, the valid values are "VIRTUAL GENERATED" or "STORED GENERATED".

Reported here: https://github.com/planetscale/cli/issues/645

internal/dumper/dumper.go
```diff
@@ -406,16 +391,15 @@ func (d *Dumper) filterDatabases(conn *Connection, filter *regexp.Regexp, invert
 	return databases, nil
 }
 
-// generatedFields returns a map that contains fields that are virtually
-// generated.
-func (d *Dumper) generatedFields(conn *Connection, table string) (map[string]bool, error) {
+// dumpableFieldNames returns a map that contains valid field names for the dump
```
fatih, line 394:
```suggestion
// dumpableFieldNames returns a slice that contains valid field names for the dump.
```

internal/dumper/dumper_test.go
```diff
@@ -14,6 +14,17 @@ import (
 	qt "github.com/frankban/quicktest"
 )
 
+func createFieldRow(name, extra string) []sqltypes.Value {
```
fatih, line 17: Usually in go we prefix variables in a test file with `test`. So this could be `testRow()`. It's read nicely and is also short.

internal/dumper/dumper.go
```diff
@@ -223,43 +223,28 @@ func (d *Dumper) dumpTable(conn *Connection, database string, table string) erro
...
 	fields := make([]string, 0)
 	{
-		cursor, err := conn.StreamFetch(fmt.Sprintf("SELECT * FROM `%s`.`%s` LIMIT 1", database, table))
+		flds, err := d.dumpableFieldNames(conn, table)
 		if err != nil {
 			return err
 		}
 
-		flds := cursor.Fields()
 		for _, fld := range flds {
```
fatih, line 233: Can be renamed to `field` or `name`

fatih (approved):
LGTM. Added a few Go style recommendations.

Issue #429, opened 2021-11-26
CLI should respect user‘s XDG base directory settings


yeskunall:
`pscale shell` creates a history file by default in `$HOME/.pscale/history`. While this is not an error, I think the CLI should respect an OS (or the user’s) XDG base directory settings. The history file should be placed under `XDG_DATA_HOME`: on Mac, it should go under `~/Library/Application \Support/.pscale` and on most Unix-like operating systems, it should go under `$HOME/.local/share`.

I’d be happy to make a PR if you guys agree with the change.

PR #672, opened 2023-05-04
Respect users XDG directory setting for history file

mscoutermarsh:
Closes: https://github.com/planetscale/cli/issues/429

This updates the CLI to use XDG_DATA_HOME for the location of the users shell history file.

The [xdg package](https://github.com/adrg/xdg) takes care of returning the path based on the operating system and creating the directory if it is not present.

## Legacy history files

For anyone with an existing `~/.pscale/history` directory, it will continue to use it. To avoid wiping out all of a person's history

### DIFF
--- a/internal/cmd/shell/shell.go
+++ b/internal/cmd/shell/shell.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/adrg/xdg"
 	"net"
 	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/planetscale/cli/internal/cmdutil"
@@ -18,7 +18,6 @@ import (
 
 	ps "github.com/planetscale/planetscale-go/planetscale"
 
-	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	exec "golang.org/x/sys/execabs"
 )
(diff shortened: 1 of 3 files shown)


Phanatic (approved):
Looks good!

internal/cmd/shell/shell.go
```diff
@@ -267,23 +266,12 @@ func formatMySQLBranch(database string, branch *ps.DatabaseBranch) string {
 }
 
 func historyFilePath(org, db, branch string) (string, error) {
-	dir, err := homedir.Dir()
+	historyFilePath := fmt.Sprintf(".pscale/history/%s.%s.%s", org, db, branch)
+
+	historyFile, err := xdg.DataFile(historyFilePath)
```
nickvanw, line 271: should we fall back to using `homedir.Dir()` if we get an error from this, just to prevent that from stopping progress?

mscoutermarsh: Thinking I might have it check if the old directory is there first so we don't wipe peoples history when this ships.

Then if not, use the new method.

nickvanw: I like that 👍

mscoutermarsh:
Pushed a couple more changes: it now checks the "legacy" folder, if it exists, continues to use it.

Otherwise uses the new location.

Also, if any of the file logic fails, it won't error. Will fallback to having no history file.

PR #708, opened 2023-07-12
root: check for updates async, and not just on errors

aybabtme:
Looks just like before, and only does the lookup/notification once per 24h period. But now the check is done async during each command invocation, with a 250ms budget affordance to finish the check if it's running.

```
A new release of pscale is available: 0.146.0 → v0.148.0
https://github.com/planetscale/cli/releases/tag/v0.148.0
```

Currently, the update check only occurs if the command ends in error! This means that there's probably a ton of `pscale` installs that are way out of date and never get prompted to update, at least until they hit an error.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -23,6 +23,7 @@ import (
 	"log"
 	"os"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


iheanyi (approved):
LGTM :shipit:

PR #725, opened 2023-08-23
Revise the phrasing of `password delete` subcmd usage to place greater emphasis on the intention of receiving a password ID

moznion:
Following the previous `password delete` subcmd's instruction, I misunderstood that I should pass the password's name instead of its ID, but actually that was wrong. The reason why I did such misoperation is I can only see the password's name on the web UI.
So this patch emphasizes the intention of receiving the password's ID on the subcmd in the usage description.


mscoutermarsh (approved):
Nice improvement, thank you!

PR #756, opened 2023-11-15
Add ability to set password `ttl` via CLI

mscoutermarsh:
The `ttl` is the time in seconds until the password will expire. A value of 0 means it will never expire.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -232,8 +232,6 @@ github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
 github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pkg/sftp v1.13.1/go.mod h1:3HaPG6Dq1ILlpPZRO0HVMrsydcdLt6HRDccSgb87qRg=
 github.com/planetscale/planetscale-go v0.51.0/go.mod h1:+rGpW2u7iQZZx4O/nFj4MZe4xIS22CVegEgl1IkTExQ=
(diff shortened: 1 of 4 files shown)


go.sum
```diff
@@ -234,6 +234,8 @@ github.com/pkg/sftp v1.13.1/go.mod h1:3HaPG6Dq1ILlpPZRO0HVMrsydcdLt6HRDccSgb87qR
 github.com/planetscale/planetscale-go v0.51.0/go.mod h1:+rGpW2u7iQZZx4O/nFj4MZe4xIS22CVegEgl1IkTExQ=
 github.com/planetscale/planetscale-go v0.92.0 h1:ifrZCSifooYuM2bG4v+RH34H2MeHQg5m+lu65EtZ9Bw=
 github.com/planetscale/planetscale-go v0.92.0/go.mod h1:fZZ8IJEB1IMrQCt3eo7D4E+hx5zt8mM3z8/aL/p1fiA=
+github.com/planetscale/planetscale-go v0.93.0 h1:cogKCS4pEbHqurZWiPROaEAo16ZEPmRTSdd2IZioa8A=
+github.com/planetscale/planetscale-go v0.93.0/go.mod h1:MR+LZLhZTIzmi2x91nwLQyIedMFhn6sgQDAKs0wAGNQ=
```
dbussink, line 238: @mscoutermarsh I think we need a `go mod tidy` here?

mscoutermarsh: 💯

internal/cmd/password/password.go
```diff
@@ -110,6 +113,7 @@ func toPasswordWithPlainText(password *ps.DatabaseBranchPassword) *PasswordWithP
 		AccessHostUrl:     password.Branch.AccessHostURL,
 		Role:              password.Role,
 		RoleDesc:          toRoleDesc(password.Role),
+		TTL:							 password.TTL,
```
dbussink, line 116: indenting seems off? Can you make sure to `go fmt` this?

dbussink (approved):

PR #759, opened 2023-11-27
Replace shell's proxy with new gRPC backed proxy

mattrobenolt:
This replaces the sql-proxy implementation with the new psdbproxy implementation.

Swapping in pscale shell is the first on the chopping block but wanted to get this integrated into the codebase first to serve as the base for the other uses of sql-proxy.

In summary, this replaces the certificate based sql-proxy, which relies on deprecated infrastructure, with a new HTTP based proxy.

This new HTTP-based proxy will be more resilient to network issues as well as acting as a connection pooler for local mysql connections.

This relies on open sourcing the `planetscale/psdbproxy` repo yet.


internal/proxyutil/psdb.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func New(cfg Config) *psdbproxy.Server {
+	return &psdbproxy.Server{
+		Addr:          cfg.LocalAddr,
+		Logger:        slog.New(zapslog.NewHandler(cfg.Logger.Core(), nil)),
+		UpstreamAddr:  cfg.UpstreamAddr,
+		Authorization: auth.NewBasicAuth(cfg.Username, cfg.Password),
+	}
+}
```
fatih, line 27: I don't think we need this extra package for constructing the server type. I would include this only if we reuse it in multiple places. Otherwise it'll just create an additional abstraction with no immediate benefits.

internal/passwordutil/password.go
```diff
@@ -0,0 +1,74 @@
...
+		prefix,
+		time.Now().Format("2006-01-02"),
+		nanoid.MustGenerate(publicIdAlphabet, publicIdLength),
+	)
+}
+
+const (
+	publicIdAlphabet = "0123456789abcdefghijklmnopqrstuvwxyz"
+	publicIdLength   = 6
+)
```
fatih, line 74: consts should be moved to the top.

internal/passwordutil/password.go
```diff
@@ -0,0 +1,74 @@
...
+				Database:     opt.Database,
+				Branch:       opt.Branch,
+				Name:         pw.Name,
+				PasswordId:   pw.PublicID,
+			})
+		},
+	}, nil
+}
+
+func GenerateName(prefix string) string {
```
fatih, line 62: Where is this used being used? Would be nice if you could add a doc comment and whether it's safe (or not) to change certain things.

internal/passwordutil/password.go
```diff
@@ -0,0 +1,74 @@
...
+	Database     string
+	Branch       string
+	Role         cmdutil.PasswordRole
+	Name         string
+	TTL          time.Duration
+}
+
+type Password struct {
+	ps.DatabaseBranchPassword
+	cleanup func(context.Context) error
```
fatih, line 25: Please let's used keyed fields. It breaks the existing struct constructions the moment you extend this struct.

fatih: > Otherwise, I think what you're suggesting is it'd need to be like,

Yeap that's what I suggested. The previous version is called "struct with unkeyed fields", hence I suggested adding field keys. 

The reason keyed structs are better is they are maintainable and easier to read. As an example, if you add a new field to the `Password` struct, if you would use unkeyed fields, it would break the compilation, and you have to refactor all existing usages of the `Password` struct. Obviously here it's just a single place, but in larger codebases it's a lot of hassle in long term. With keyed fields, adding a new field doesn't break existing usages, and everything works as it is, and no further changes are required.

fatih (approved):

internal/passwordutil/password.go
mattrobenolt: Yeah, I understand the differences and made a conscious decision on my choice for the ergonomics that I felt were better. I made the change anyways as you preferred, I just disagree that it's better that way, but it's bike shedding. I also chose to embed the struct to not have another pointer dereference, but that's an extreme nit on my part. Just something tangibly different in the implementations.

internal/proxyutil/psdb.go
mattrobenolt: For context, #760 #761 and #762 are the other implementations of this.

PR #765, opened 2023-11-28
Add new `ping` command

mattrobenolt:
Depends on https://github.com/planetscale/planetscale-go/pull/186

I'm not sure what all goes into adding a new command, but I assume documentation and whatnot.

I'm also very open to how to display this information, but right now, we get a table like:

```
  ENDPOINT (17)                                    LATENCY   TYPE
 ------------------------------------------------ --------- -----------
  aws.connect.psdb.cloud                           32.9ms    optimized
  us-west.connect.psdb.cloud                       35.7ms    direct
  gcp-us-central1.connect.psdb.cloud               57.3ms    direct
  gcp.connect.psdb.cloud                           59ms      optimized
  aws-us-east-2.connect.psdb.cloud                 60.4ms    direct
  gcp-us-east4.connect.psdb.cloud                  68.5ms    direct
  us-east.connect.psdb.cloud                       71ms      direct
  gcp-northamerica-northeast1.connect.psdb.cloud   83.4ms    direct
  ap-northeast.connect.psdb.cloud                  122.4ms   direct
  aws-ap-southeast-2.connect.psdb.cloud            146.9ms   direct
  gcp-asia-northeast3.connect.psdb.cloud           153.9ms   direct
  eu-central.connect.psdb.cloud                    155.1ms   direct
  aws-eu-west-2.connect.psdb.cloud                 163.5ms   direct
  eu-west.connect.psdb.cloud                       173.9ms   direct
  aws-sa-east-1.connect.psdb.cloud                 182.1ms   direct
  ap-southeast.connect.psdb.cloud                  195.8ms   direct
  ap-south.connect.psdb.cloud                      238.1ms   direct
```

I didn't really like having them all combined into 1 table, I wanted to split optimized vs direct into different groups, but our CLI's formatting definitely expects to return some JSON object for the results, so figured adding a "type" field was the best option.

But this should be able to help folks quickly diagnose what the best endpoint is for them while running from an actual VM or wherever.

There are a couple use cases for this tool:

* help debug latency issues, if the "optimized" endpoint appears to be resolving wrong, this could identify where it's going wrong and if you should prefer the "direct"
* If you're hosted in a provider that isn't AWS or GCP, you might not exactly know which of our regions are the closest to your server, so this tool identify the closest region.


iheanyi:
I think this is a helpful feature, but I don't think `ping` should be a top-level command because there will be no subcommands for it. How would you feel about adding it under something like a `debug` command. This also gives us room for adding more subcommands under the `debug` command as necessary. What do you think?

mattrobenolt:
I have no opinions, I'm fine with that.

iheanyi (approved):
Talked it out a bit more offline, `pscale ping` works.

PR #790, opened 2023-12-19
PSDBProxy take 2

dbussink:
This adds the changes to move to `psdbproxy` again now that we've significantly improved the Vitess dependency situation and we can cross compile to Windows as well.


internal/cmd/database/dump.go
```diff
@@ -141,10 +120,57 @@ func dump(ch *cmdutil.Helper, cmd *cobra.Command, flags *dumpFlags, args []strin
...
+	}()
+
+	localAddr := "127.0.0.1:0"
+	if flags.localAddr != "" {
+		localAddr = flags.localAddr
+	}
+
+	remoteAddr := flags.remoteAddr
+	if remoteAddr == "" {
+		remoteAddr = pw.Password.Hostname
```
fatih, line 150: The description says that remote addr is in format `host:port`, but here we only use the hostname. Does the hostname also include the port?

fatih (approved):

internal/cmd/database/dump.go
dbussink: It doesn't have it. It falls back to `443` as the default if no port is given.

PR #805, opened 2024-01-08
internal/update: remove test error field

mdlayher:
I had a slight variation of this before but removed it due to an unexpected actions failure. This should cover our bases so we can report something useful.


mdlayher:
Welp, appears the actions runner is basically root. I'm gonna remove this field entirely.

mscoutermarsh (approved):

PR #762, opened 2023-11-27
connect: port to using psdbproxy

mattrobenolt:

PR #817, opened 2024-01-26
cmd/connect: replace connect with psdbproxy

mattrobenolt:
Newer version of https://github.com/planetscale/cli/pull/762 now with credential renewing so it should work for long running situations.

As a follow-up, I would like to add the ability to use static username/password credentials rather than generating ephemeral credentials.

I think that'll open up the door for more production capable use cases.

All around, having `pscale connect` work through here should start becoming a better path for production workloads since it now acts as a local connection pooler.


internal/cmd/connect/connect.go
```diff
@@ -108,58 +109,88 @@ argument:
...
+			})
+			if err != nil {
+				return cmdutil.HandleError(err)
+			}
+			defer func() {
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+				defer cancel()
 
-			localAddr := net.JoinHostPort(flags.host, flags.port)
+				if err := pw.Cleanup(ctx); err != nil {
```
notfelineit, line 127: Nice Cleanup()

notfelineit (approved):

dbussink (approved):

PR #819, opened 2024-01-29
cmd/password: use a duration for TTL

mdlayher:
Go has a native duration type and it's much friendlier than requiring the caller to do integer math. `--ttl 6h`, `--ttl 1h30m`, etc.

To prevent misuse, require the caller to specify _at least_ 1 second as the TTL when the flag is set.

### DIFF
--- a/internal/cmd/password/create.go
+++ b/internal/cmd/password/create.go
@@ -3,6 +3,7 @@ package password
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/planetscale/cli/internal/cmdutil"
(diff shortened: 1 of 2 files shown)


mattrobenolt:
I'm not sure I agree with this in this context. This definitely opens the door to attempting to use fractional seconds. I know it's being rounded, but trying to use 5500ms or something is just going to be wrong.

I think in the context of a TTL being seconds, this is a very common format outside of Go.

This is also a breaking change since someone who maybe had a script doing --ttl=30 will now fail.

mdlayher:
As an option, we can easily prevent fractional seconds in the conversion function.

We can also make the flag a string and infer that `--ttl 30` is 30 seconds, while still also recognizing `--ttl 30s`.

mattrobenolt:
Sure. I'm doing to defer to others for this. I don't feel that it's worth it or solves anything.

mdlayher:
The latest commit preserves the ability to pass bare integers in seconds while also allowing Go duration strings such as `30m`. No sub-second rounding is permitted.

mattrobenolt (approved):
Ok, yeah, I like this better now. Seems like no downsides.

PR #823, opened 2024-01-29
meta: move goreleaser-cross back to v1.21.5

mattrobenolt:
Turns out, these are different releases for goreleaser-cross vs goreleaser, and v1.21.5 is the latest for -cross variant.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ endif
 REPO=planetscale
 NAME=pscale
 BUILD_PKG=github.com/planetscale/cli/cmd/pscale
-GORELEASE_CROSS_VERSION ?= v1.23.0
+GORELEASE_CROSS_VERSION ?= v1.21.5
 SYFT_VERSION ?= 0.102.0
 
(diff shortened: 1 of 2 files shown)


iheanyi (approved):
Wasn't ours before at 1.21.6? Was that also wrong?

mattrobenolt:
Ok, so in summary, the last build that worked was with 1.21.5. I think there are a few things that are worth investigating:

* there is no goreleaser-cross tag for 1.21.6. From what I'm deriving, this version _is_ the same as the go version. So it might be an oversight in their repo to not have a tag exist for 1.21.6 while it's published to docker registries.
* The 1.21.6 tag also brings goreleaser 1.30.0, whereas 1.21.5 tag brings 1.22.1. I suspect more the issue is with the goreleaser bump. afaik there's not a goreleaser-cross:v1.21.6-v1.22.1 that keeps the older goreleaser to test with.

Given that the last successful publish was on the v1.21.5 tag, I'm inclined to try this.

PR #838, opened 2024-03-23
Add update-vschema command

nickvanw:
Depends on https://github.com/planetscale/planetscale-go/pull/198

Implements a `pscale branch update-vschema`, which takes either a file to read from or the new vschema via standard in.

Experimental feature, if you're reading this, contact support if you're interested!


nickvanw:
I think I'll probably update this so that it is a subcommand of `vschema`, I like that better. We'll have a high-level one for routing rules as well.

dbussink (approved):

PR #844, opened 2024-04-01
Add `--replica` flag to create password

notfelineit:
This PR adds `--replica` flag to create password, which when enabled, will create a password that redirects queries to the replica tablets, or the  branch's read-only-regions, depending on where latency is smallest.

### DIFF
--- a/internal/cmd/password/create.go
+++ b/internal/cmd/password/create.go
@@ -16,8 +16,9 @@ import (
 
 func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 	var flags struct {
-		role string
-		ttl  ttlFlag
+		role    string
+		ttl     ttlFlag
+		replica bool
 	}
 
 	cmd := &cobra.Command{
@@ -52,6 +53,7 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 				Name:         name,
 				Role:         flags.role,
 				TTL:          int(flags.ttl.Value.Seconds()),
+				Replica:      flags.replica,
 			})
 			if err != nil {
 				switch cmdutil.ErrCode(err) {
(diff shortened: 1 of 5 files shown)


internal/cmd/password/create.go
```diff
@@ -76,6 +78,8 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 	cmd.PersistentFlags().StringVar(&flags.role, "role",
 		"admin", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is admin.")
 	cmd.PersistentFlags().Var(&flags.ttl, "ttl", `TTL defines the time to live for the password. Durations such as "30m", "24h", or bare integers such as "3600" (seconds) are accepted. The default TTL is 0s, which means the password will never expire.`)
+	cmd.Flags().BoolVar(&flags.replica, "replica", false, "When enabled, the password will route all reads to the branch's primary replicas, and all read-only regions.")
+	cmd.Flags().MarkHidden("replica")
```
notfelineit, line 82: We may be able to remove this.

internal/cmd/password/create.go
```diff
@@ -76,6 +78,8 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 	cmd.PersistentFlags().StringVar(&flags.role, "role",
 		"admin", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is admin.")
 	cmd.PersistentFlags().Var(&flags.ttl, "ttl", `TTL defines the time to live for the password. Durations such as "30m", "24h", or bare integers such as "3600" (seconds) are accepted. The default TTL is 0s, which means the password will never expire.`)
+	cmd.Flags().BoolVar(&flags.replica, "replica", false, "When enabled, the password will route all reads to the branch's primary replicas, and all read-only regions.")
```
iheanyi, line 81:
```suggestion
	cmd.Flags().BoolVar(&flags.replica, "replica", false, "When enabled, the password will route all reads to the branch's primary replicas and all read-only regions.")
```

internal/cmd/password/password.go
```diff
@@ -107,6 +110,7 @@ func toPassword(password *ps.DatabaseBranchPassword) *Password {
 		Username:  password.Username,
 		Role:      password.Role,
 		RoleDesc:  toRoleDesc(password.Role),
+		Replica:   password.Replica,
```
iheanyi, line 113: I wonder if we should do something like `Connection type: Primary / Replica` here instead of a boolean?

iheanyi (approved):

internal/cmd/password/password.go
notfelineit: Ooh, yeah!

iheanyi (approved):

PR #845, opened 2024-04-01
Default to `reader` role when passing `--replica` and not `--role`

notfelineit:
This PR makes it so that we default to a `reader` role when passing `--replica`, and not `--role`. 
- The default `admin` for `--replica` flag is set in the `RunE` func instead of the flag definition
    - This is so we can differentiate between a user passing no `--role`, and the user intentionally setting `--role=admin`
- If no `--role` is passed, we default it to `reader` if `--replica` is passed, and `admin` if `--replica` is _not_ passed

### DIFF
--- a/internal/cmd/password/create.go
+++ b/internal/cmd/password/create.go
@@ -31,9 +31,13 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 			branch := args[1]
 			name := args[2]
 
-			// "admin" is the default value for role
-			if flags.role == "" && flags.replica {
-				flags.role = "reader"
+			if flags.role == "" {
+				if flags.replica {
+					flags.role = "reader"
+				} else {
+					// Maintain old behavior - "admin" is the default role.
+					flags.role = "admin"
+				}
 			}
 
 			if flags.role != "" {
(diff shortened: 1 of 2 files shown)


dgraham (approved):

iheanyi (approved):

internal/cmd/password/create.go
```diff
@@ -76,7 +85,7 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 		},
 	}
 	cmd.PersistentFlags().StringVar(&flags.role, "role",
-		"admin", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is admin.")
+		"", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is admin.")
```
iheanyi, line 88: Can we update the copy here?

```suggestion
		"", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is admin.")
```
```suggestion
		"", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. Defaults to 'reader' for replicas, otherwise defaults to 'admin'.")
```

iheanyi, line 88:
```suggestion
		"", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. Defaults to 'reader' for replica passwords, otherwise defaults to 'admin'.")
```

PR #846, opened 2024-04-03
Add support for replicas in `shell` and `connect` commands

iheanyi:
This pull request implements replica credential creation inside of both the shell and connect commands.

### DIFF
--- a/internal/cmd/connect/connect.go
+++ b/internal/cmd/connect/connect.go
@@ -33,6 +33,7 @@ func ConnectCmd(ch *cmdutil.Helper) *cobra.Command {
 		execCommandEnvURL   string
 		role                string
 		noRandom            bool
+		replica             bool
 	}
 
 	cmd := &cobra.Command{
@@ -82,12 +83,16 @@ argument:
 				}
 			}
 
+			replica := flags.replica
+
 			role := cmdutil.AdministratorRole
 			if flags.role != "" {
 				role, err = cmdutil.RoleFromString(flags.role)
 				if err != nil {
 					return err
 				}
+			} else if replica {
+				role = cmdutil.ReaderRole
 			}
 
 			// check whether database and branch exist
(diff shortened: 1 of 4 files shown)


internal/cmd/shell/shell.go
```diff
@@ -182,7 +188,11 @@ second argument:
 				"-t", // the -s (silent) flag disables tabular output, re-enable it.
 				"-h", host,
 				"-P", port,
-				"-D", "@primary",
+			}
+			if replica {
+				mysqlArgs = append([]string{"--no-defaults"}, mysqlArgs...)
```
mattrobenolt, line 193: Is there a reason you chose to _prepend_ `--no-defaults`?

mattrobenolt (approved):

internal/cmd/connect/connect.go
```diff
@@ -207,7 +213,9 @@ argument:
 	cmd.PersistentFlags().StringVar(&flags.execCommandEnvURL, "execute-env-url", "DATABASE_URL",
 		"Environment variable name that contains the exposed Database URL.")
 	cmd.PersistentFlags().StringVar(&flags.role, "role",
-		"admin", "Role defines the access level, allowed values are : reader, writer, readwriter, admin. By default it is admin.")
+		"", "Role defines the access level, allowed values are: reader, writer, readwriter, admin. Defaults to 'reader' for replica passwords, otherwise defaults to 'admin'.")
```
notfelineit, line 216: Since we took out the default `admin` here, should we check for empty string above and set to admin if not replica? (to preserve old functionality)

internal/cmd/shell/shell.go
iheanyi: Because it didn't work otherwise 🤷🏾

internal/cmd/connect/connect.go
iheanyi: @notfelineit Yeah, we do that by default already. The role is set to `admin` by default in the code and if the replica flag is enabled, we override it to be `reader`.

notfelineit: Wow. I'm blind cause I totally missed that!

Issue #850, opened 2024-04-06
[Feature Request] Do not disconnect `pscale` CLI on CTRL+C


twiddler:
In the CLI, I often find myself wanting to clear a command I was entering with <kbd>Ctrl</kbd>+<kbd>C</kbd>. That leads to either a disconnect or my terminal hanging in a weird limbo state. It'd be nice if it just cleared the current command I was about to send, and let me enter a new one.

(I would be perfectly fine with having to hit <kbd>Ctrl</kbd>+<kbd>C</kbd> twice to disconnect in those cases that I already started typing something.)

twiddler:
The new behaviour seems to be broken:

1. I enter a query (`select foo`) without submitting. I hit <kbd>Ctrl</kbd>+<kbd>C</kbd>. `^C` appears and I can enter a new query. So far, so good.
2. I enter a new query (`select count(*) from SomeTable`) and hit <kbd>↵ Enter</kbd> to submit it. `^C` appears again, and my query (apparently) doesn't get submitted. This is wrong; I expected my query to be submitted.
3. I enter the query again and submit it. I get a response.

```console
$ pscale shell org branch
org/branch> select foo^C
org/branch> select count(*) from SomeTable;
^C
org/branch> select count(*) from SomeTable;
+----------+
| count(*) |
+----------+
|        1 |
+----------+
org/branch> 
```

<kbd>Ctrl</kbd>+<kbd>D</kbd> does work, although it takes up to a few seconds and sometimes produces some errors:

```console
org/branch> ^Dfailed to delete credentials:  Delete "https://api.planetscale.com/v1/organizations/account/databases/org/branches/branch/passwords/asdfgh123456": context deadline exceeded
Error: os: process already finished
```

Versions: 

- pscale version 0.195.0 (build date: 2024-04-18T15:27:49Z commit: e37d329)
- bash version 5.2.26(1)-release
- GNOME Terminal version 3.48.3 for GNOME 44
- Fedora 38

dbussink:
> 2\. I enter a new query (`select count(*) from SomeTable`) and hit ↵ Enter to submit it. `^C` appears again, and my query (apparently) doesn't get submitted. This is wrong; I expected my query to be submitted.

For some reason I'm not sure about yet, Go seems to double deliver the signals which is why this seems to be happening. I see the interrupt being delivered twice instead of once, so I have to find out why that's the case. That's the cause of this issue, since the second one is then also interrupted due to the double delivery. 

[...]

PR #854, opened 2024-04-22
Redo signal handling

dbussink:
These changes rework how we do signal handling after #853. It turns out, trying to forward signals is not really what we want.

At the code here is that different platforms behave in different ways with how signals are handled. The specific cause of the remaining issue described in #850 is that on Linux and unix like platforms, the normal behavior on `ctrl+c` is that all processes in the current group get sent a `ctrl+c`.

If we forward signals, we receive the `ctrl+c` twice inside MySQL leading to weird behavior. But we also don't want to ignore the signals either, since a single `SIGINT` sent to `pscale` is still something we want to act on. On Windows though, the current signal handling setup does seem to work as desired.

So we have to split things up here per platform. On Unix like platforms, we can use https://pkg.go.dev/syscall#SysProcAttr and set `Foreground`. That will imply `Setpgid` as well, which means we get a new process group for the process we're spawning and it attaches it to the foreground. That means now MySQL has the foreground as an independent process group, so it gets the signals appropriately and it's as if we'd directly have started MySQL. 

On Windows we keep using the same signal handling logic mostly, but slightly simplified. Lastly, we fix the warning signal for prod branches since on Windows it looks like `⚠` doesn't actually work.

### DIFF
--- a/internal/cmd/shell/shell.go
+++ b/internal/cmd/shell/shell.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"time"
 
@@ -23,7 +22,7 @@ import (
 	"github.com/planetscale/cli/internal/proxyutil"
 )
 
-func ShellCmd(ch *cmdutil.Helper) *cobra.Command {
+func ShellCmd(ch *cmdutil.Helper, sigc chan os.Signal, signals ...os.Signal) *cobra.Command {
 	var flags struct {
 		localAddr  string
 		remoteAddr string
@@ -213,7 +212,7 @@ second argument:
 			}()
 
 			go func() {
-				errCh <- m.Run(ctx, mysqlArgs...)
+				errCh <- m.Run(ctx, sigc, signals, mysqlArgs...)
 			}()
 
 			go func() {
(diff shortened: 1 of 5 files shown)


internal/cmd/shell/shell.go
```diff
@@ -272,58 +271,17 @@ func (m *mysql) Run(ctx context.Context, sigc chan os.Signal, signals []os.Signa
...
-	// When we exit from this function, we restore the signals as they were.
-	signal.Stop(sigc)
-	defer signal.Notify(sigc, signals...)
-
-	err := c.Start()
-	if err != nil {
-		return err
-	}
+	c.SysProcAttr = sysProcAttr()
+	setupSignals(ctx, c, sigc, signals)
```
fatih, line 275: Why is the return value not used? I believe it's non-nil for Windows.

internal/cmd/shell/shell_windows.go
```diff
@@ -0,0 +1,58 @@
...
+	// When we exit from this function, we restore the signals as they were.
+	signal.Stop(sigc)
+
+	go func() {
+		for {
+			select {
+			case sig := <-msig:
+				_ = c.Process.Signal(sig)
+			case <-ctx.Done():
+				return
```
fatih, line 48: Should we call `signal.Stop(msig)` here if the ctx is cancelled?

fatih (approved):

internal/cmd/shell/shell.go
dbussink: Good catch, let me fix this up.

internal/cmd/shell/shell_windows.go
dbussink: No, this is what the caller should take care of (but what was missing, see your other question :smile:).

fatih: Yeah, I wasn't sure who was handling the stop call :)

fatih (approved):

PR #883, opened 2024-07-11
Fix postitive auth check message

jgreet:
Fixes a missing newline in the affirmative `pscale auth check` output

Currently:
```sh
bash-3.2$ pscale auth check
You are authenticated.bash-3.2$
```

### DIFF
--- a/internal/cmd/auth/check.go
+++ b/internal/cmd/auth/check.go
@@ -38,7 +38,7 @@ func CheckCmd(ch *cmdutil.Helper) *cobra.Command {
 						ExitCode: cmdutil.ActionRequestedExitCode,
 					}
 				} else {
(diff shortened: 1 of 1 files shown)


internal/cmd/auth/check.go
```diff
@@ -38,7 +38,7 @@ func CheckCmd(ch *cmdutil.Helper) *cobra.Command {
 						ExitCode: cmdutil.ActionRequestedExitCode,
 					}
 				} else {
-					ch.Printer.Printf("You are authenticated.")
+					ch.Printer.Printf("You are authenticated.\n")
```
nickvanw, line 41:
```suggestion
					ch.Printer.Println("You are authenticated.")
```

PR #884, opened 2024-07-11
Make region and restore flags mutually exclusive for branch create

### LINKED ISSUE #874: `pscale branch create` ignores `--region` when `--restore` is used
I'm using `pscale branch create` to restore a backup to a new branch, ideally in a different region.

If I just specify the region (`--region`), the branch is created in the correct location:

```sh
# Works as expected ✅
pscale branch create platform my-new-branch --from main --region aws-ap-southeast-2
```

If I just specify the backup id (`--restore`), the correct data is restored and the branch is created in the parent branches region, as expected (`us-west-2` in this case):

```sh
# Works as expected ✅
pscale branch create platform my-new-branch --from main --restore EtvjZWF4O1zw
```

But if I specify both `--restore` and `--region`, the region argument seems to be ignored:

```sh
# Ignores the region? 🫤
pscale branch create platform my-new-branch --from main --region aws-ap-southeast-2 --restore EtvjZWF4O1zw
```

At a technical level I can understand why this would occur but it being able to do this would be pretty handy. If the current behaviour is intended, it should at least be mentioned in [the docs](https://planetscale.com/docs/reference/branch) and the cli should possibly print a warning when run under these conditions.

jgreet:
A fairly common point of confusion is that `pscale branch create --restore` will accept a `--region` flag but will always restore to the region the backup was taken in. That's the correct behavior but it isn't documented properly.

This changes makes those flags mutually exclusive and documents the incompatibility in the app.

```sh
$ pscale branch create beam main --region us-west --restore abc123
Error: if any flags in the group [region restore] are set none of the others can be; [region restore] were all set
```

```sh
$ pscale branch create --help
Create a new branch from a database
...
Flags:
      --from string      Branch to be created from
  -h, --help             help for create
      --region string    Region for the database. Can not be used with --restore
      --restore string   Backup to restore into the branch. Backup can only be restored to its original region
      --seed-data        Add seed data using the Data Branching™ feature
      --wait             Wait until the branch is ready
...
```

A potential issue with this change is that that providing `--restore` and `--region` with the backup's correct region currently works as expected but would fail with this merged.

Fixes #874

### DIFF
--- a/internal/cmd/branch/create.go
+++ b/internal/cmd/branch/create.go
@@ -118,11 +118,12 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&createReq.ParentBranch, "from", "", "branch to be created from")
-	cmd.Flags().StringVar(&createReq.Region, "region", "", "region for the database")
-	cmd.Flags().StringVar(&createReq.BackupID, "restore", "", "backup to restore into the branch")
-	cmd.Flags().BoolVar(&flags.dataBranching, "seed-data", false, "add seed data using the Data Branching™ feature")
-	cmd.Flags().BoolVar(&flags.wait, "wait", false, "wait until the branch is ready")
+	cmd.Flags().StringVar(&createReq.ParentBranch, "from", "", "Branch to be created from")
+	cmd.Flags().StringVar(&createReq.Region, "region", "", "Region for the database. Can not be used with --restore")
+	cmd.Flags().StringVar(&createReq.BackupID, "restore", "", "Backup to restore into the branch. Backup can only be restored to its original region")
+	cmd.Flags().BoolVar(&flags.dataBranching, "seed-data", false, "Add seed data using the Data Branching™ feature")
+	cmd.Flags().BoolVar(&flags.wait, "wait", false, "Wait until the branch is ready")
+	cmd.MarkFlagsMutuallyExclusive("region", "restore")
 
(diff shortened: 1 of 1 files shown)


internal/cmd/branch/create.go
```diff
@@ -118,11 +118,12 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
-	cmd.Flags().StringVar(&createReq.Region, "region", "", "region for the database")
-	cmd.Flags().StringVar(&createReq.BackupID, "restore", "", "backup to restore into the branch")
-	cmd.Flags().BoolVar(&flags.dataBranching, "seed-data", false, "add seed data using the Data Branching™ feature")
-	cmd.Flags().BoolVar(&flags.wait, "wait", false, "wait until the branch is ready")
+	cmd.Flags().StringVar(&createReq.ParentBranch, "from", "", "Branch to be created from")
+	cmd.Flags().StringVar(&createReq.Region, "region", "", "Region for the database. Can not be used with --restore")
+	cmd.Flags().StringVar(&createReq.BackupID, "restore", "", "Backup to restore into the branch. Backup can only be restored to its original region")
+	cmd.Flags().BoolVar(&flags.dataBranching, "seed-data", false, "Add seed data using the Data Branching™ feature")
+	cmd.Flags().BoolVar(&flags.wait, "wait", false, "Wait until the branch is ready")
+	cmd.MarkFlagsMutuallyExclusive("region", "restore")
```
nickvanw, line 126: TIL, I like this

nickvanw (approved):

PR #909, opened 2024-09-27
Add `keyspace create` command

iheanyi:
This pull request allows users to create a keyspace via the CLI.


internal/cmd/keyspace/create.go
```diff
@@ -0,0 +1,76 @@
...
+				ch.Printer.Printf("Keyspace %s was successfully created.\n", printer.BoldBlue(k.Name))
+				return nil
+			}
+
+			return ch.Printer.PrintResource(toBranchKeyspace(k))
+		},
+	}
+	cmd.Flags().IntVar(&flags.shards, "shards", 1, "Number of shards in the keyspace")
+	cmd.Flags().StringVar(&flags.clusterSize, "cluster-size", "PS_10", "cluster size for the keyspace. Options: PS_10, PS_20, PS_40, PS_80, PS_160, PS_320, PS_400")
+	cmd.Flags().IntVar(&flags.replicas, "replicas", 0, "Number of additional replicas per shard. By default, each cluster comes with 2 replicas.")
```
nickvanw, line 67:
```suggestion
	cmd.Flags().IntVar(&flags.replicas, "replicas", 0, "Number of additional replicas per shard. By default, each cluster comes with 2 replicas.")
```

I think this should probably be `additional-replicas`

notfelineit (approved):

Issue #540, opened 2022-07-28
Improve restore-dump flexibility to allow simpler restores into a differently named database


orware:
Currently, the `restore-dump` subcommand will assume that you are attempting to always restore into a database that has the same name as the `<DATABASE_NAME>` prefix that is added into the generated files from the `dump` subcommand.

The generated files have a structure that looks like this:

```
<DATABASE_NAME>.<TABLE_NAME>-schema.sql 
<DATABASE_NAME>.<TABLE_NAME>.00001.sql
...
```

For example, let's say I have a database named `old-name`.

This will generate files that have that prefix.

```
old-name.example-table-schema.sql
old-name.example-table.00001.sql
```

This works perfectly fine for `restore-dump`, so long as the database name you are attempting to restore into matches the filenames, which would be `old-name` in this case.

But let's say the old database name actually was wrong, or there was some need to change it over time, so now you want it to be called `new-name`.

If you attempted to run a command like this:

```
pscale database restore-dump new-name main --dir "/planetscale-dumps/old-name" --org=example
```

You would run into the following error, since the `restore-dump` command looks at the filenames within the dump folder and sees the `old-name` prefix:

```
Error: failed to restore database: unknown database 'old-name' (errno 1105) (sqlstate HY000)
```

The current workaround is to perform an extra step where you need to rename the files in the dump folder to have the new database name like this:

```
new-name.example-table-schema.sql
new-name.example-table.00001.sql
```

It would be nice to not have to do this however, particularly because in the `restore-dump` example above, we are already providing `new-name` as part of the input parameters.

Having an additional flag that would allow the embedded `<DATABASE_NAME>` in the filenames to be ignored and automatically use the provided database name passed to `restore-dump` as an input parameter would be helpful if it were available and would allow for easier reuse of the dump folders that are generated for different databases.

I had planned to look into this myself at some point and attempt a PR, but if someone tackles it more quickly then that's perfectly fine too.

aaacafe786:
that is really awesome, thank you all for your hard work

orware:
Addressed by https://github.com/planetscale/cli/pull/910 once that is able to get merged in via the new `--allow-different-destination` flag.

PR #910, opened 2024-09-27
Add optional progress information and other enhancements for restore-dump command

orware:
These changes add a few potentially useful capabilities to the `restore-dump` command.

The first is the `--show-details` flag, which helps provide a middle ground between the minimal default "Restoring database ..." output and the potentially too noisy `--debug` output, since that ends up outputting the large `INSERT` queries being executed as well to the terminal. This flag will output the names of the files that will be included within the restore folder and will also show progress processing the queries within the various data files.

~~The `--start-from` flag allows you to provide a table name which will then be used to skip earlier tables so that the restore will begin from the table name that is provided which can be helpful to avoid having to start a restore completely from the very first table again.~~
(Replaced by the `--starting-table` and `--ending-table` options described in the later comment down below.)

The `--allow-different-destination` flag addresses https://github.com/planetscale/cli/issues/540 and adds extra flexibility since users would not be required to adjust the database name prefix embedded into the files within a dump folder for a different database and can simply enable this option and it will allow those files to be utilized for a database with a different name.

The `--schema-only` and `--data-only` flags can help address situations where you may have a folder containing both sets of files but only need one type. While a new folder containing only the files needed is an option this mainly helps add extra flexibility without needing to take that extra step.

### Example output:

The following two examples use an example database name of `metrics` into its `main` branch.

The first example shows using the `--schema-only` flag:
```
pscale database restore-dump metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details

Starting to restore database metrics from folder <DIRECTORY_NAME>
Restoring database ...
The schema only option is enabled for this restore.
Collecting files from folder <DIRECTORY_NAME>
  |- Table file: metrics.connections-schema.sql
    |- Data file: metrics.connections.00001.sql
    |- Data file: metrics.connections.00002.sql
    |- Data file: metrics.connections.00003.sql
    |- Data file: metrics.connections.00004.sql
    |- Data file: metrics.connections.00005.sql
  |- Table file: metrics.placeholder-schema.sql
Dropping Existing Table (if it exists): `metrics`.`connections`
Creating Table: `metrics`.`connections` (Table 1 of 2)
Dropping Existing Table (if it exists): `metrics`.`placeholder`
Creating Table: `metrics`.`placeholder` (Table 2 of 2)
Skipping restoring data files...
Restore is finished! (elapsed time: 1.189019208s)
```

The second example demonstrates how using `--start-from` prevents files from the earlier `connections` table from being included at all and allows things to start from the provided table name which happens to be called `placeholder`:

```
pscale database restore-dump metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details --start-from="placeholder"

Starting to restore database metrics from folder <DIRECTORY_NAME>
Restoring database ...
The schema only option is enabled for this restore.
Collecting files from folder <DIRECTORY_NAME>
Skipping files associated with the connections table...
  |- Table file: metrics.placeholder-schema.sql
Starting from placeholder table...
Dropping Existing Table (if it exists): `metrics`.`placeholder`
Creating Table: `metrics`.`placeholder` (Table 1 of 1)
Skipping restoring data files...
Restore is finished! (elapsed time: 562.755959ms)
```

To demonstrate the usage of `--allow-different-destination` first let's see the default behavior when attempting to restore a dump folder, created for the `metrics` database, into a database named `not-metrics`:

```
pscale database restore-dump not-metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details --start-from="placeholder"

Starting to restore database not-metrics from folder <DIRECTORY_NAME>
Restoring database ...
The schema only option is enabled for this restore.
Collecting files from folder <DIRECTORY_NAME>
Skipping files associated with the connections table...
  |- Table file: metrics.placeholder-schema.sql
Starting from placeholder table...
Error: failed to restore database: VT05003: unknown database 'metrics' in vschema (errno 1105) (sqlstate HY000)
```

As expected, the restore is unable to complete since the filenames within the dump folder have the embedded `metrics` name which doesn't match with the `not-metrics` name being passed into the command.

But if we add the new `--allow-different-destination` flag in, this allows the restore to proceed into the differently named destination database.

```
pscale database restore-dump not-metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details --start-from="placeholder" --allow-different-destination

Starting to restore database not-metrics from folder <DIRECTORY_NAME>
Restoring database ...
The allow different destination option is enabled for this restore.
Files that do not begin with the provided database name of not-metrics will still be processed without having to rename them first.
The schema only option is enabled for this restore.
Collecting files from folder <DIRECTORY_NAME>
Skipping files associated with the connections table...
  |- Table file: metrics.placeholder-schema.sql
Starting from placeholder table...
Dropping Existing Table (if it exists): `not-metrics`.`placeholder`
Creating Table: `not-metrics`.`placeholder` (Table 1 of 1)
Skipping restoring data files...
Restore is finished! (elapsed time: 543.931875ms)
```

I didn't want to include a lot of lines showing data files being processed but below is a snippet showing the extra details that would be included towards the end of processing a file:

```
  Processing Query 126 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 127 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 128 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 129 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 130 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 131 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 132 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 133 out of 134 within metrics.connections.00005.sql in thread 1
  Processing Query 134 out of 134 within metrics.connections.00005.sql in thread 1
Finished Processing Data File: metrics.connections.00005.sql in 2m45.760835334s with 6m20.652604042s elapsed so far (File 5 of 5)
```


orware:
I was thinking about this some more and it makes sense to allow for this option to be more comprehensive and allow not just >= a starting table to be included but a <= an ending table too or being able to provide a more limited starting/ending table range so I made some adjustments.

The last commit removes the `--start-from` option and replaces it with the `--starting-table` and `--ending-table` options which can be used separately or together.

### Examples:

If not using the options at all then all tables should be included as they normally would. If the ending table provided comes before the starting table alphabetically then the restore will short circuit until that is corrected.

This example skips the last two tables since the table range being provided will only include the first two:
```
pscale database restore-dump metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details --starting-table="abc" --ending-table="bcd"

Starting to restore database metrics from folder <DIRECTORY_NAME>
Restoring database ...
The schema only option is enabled for this restore.
Collecting files from folder <DIRECTORY_NAME>
  |- Table file: metrics.abc-schema.sql
  |- Table file: metrics.bcd-schema.sql
Skipping files associated with the connections table...
Skipping files associated with the placeholder table...
Restore will be starting from the abc table...
Restore will be ending at the bcd table...
Dropping Existing Table (if it exists): `metrics`.`abc`
Creating Table: `metrics`.`abc` (Table 1 of 2)
Dropping Existing Table (if it exists): `metrics`.`bcd`
Creating Table: `metrics`.`bcd` (Table 2 of 2)
Skipping restoring data files...
Restore is finished! (elapsed time: 1.005391166s)
```

For the case above, running the same command without the use of `--starting-table` would result in the same output since the `abc` table is the only other one before `bcd`:

```
pscale database restore-dump metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details --ending-table="bcd"
```

The opposite however, starting from the `bcd` table and continuing to the end will include the remaining tables automatically:

```
database restore-dump metrics main --dir="<DIRECTORY_NAME>" --org=<ORG> --overwrite-tables --threads=2 --schema-only --show-details --starting-table="bcd"

Starting to restore database metrics from folder <DIRECTORY_NAME>
Restoring database ...
The schema only option is enabled for this restore.
Collecting files from folder <DIRECTORY_NAME>
Skipping files associated with the abc table...
  |- Table file: metrics.bcd-schema.sql
  |- Table file: metrics.connections-schema.sql
    |- Data file: metrics.connections.00001.sql
    |- Data file: metrics.connections.00002.sql
    |- Data file: metrics.connections.00003.sql
    |- Data file: metrics.connections.00004.sql
    |- Data file: metrics.connections.00005.sql
  |- Table file: metrics.placeholder-schema.sql
Restore will be starting from the bcd table...
Dropping Existing Table (if it exists): `metrics`.`bcd`
Creating Table: `metrics`.`bcd` (Table 1 of 3)
Dropping Existing Table (if it exists): `metrics`.`connections`
Creating Table: `metrics`.`connections` (Table 2 of 3)
Dropping Existing Table (if it exists): `metrics`.`placeholder`
Creating Table: `metrics`.`placeholder` (Table 3 of 3)
Skipping restoring data files...
Restore is finished! (elapsed time: 1.321406s)
```

internal/dumper/pool.go
```diff
@@ -40,7 +40,7 @@ func (conn *Connection) StreamFetch(query string) (driver.Rows, error) {
 // NewPool creates the new pool.
 func NewPool(log *zap.Logger, cap int, address string, user string, password string, vars []string, database string) (*Pool, error) {
 	conns := make(chan *Connection, cap)
-	for i := 0; i < cap; i++ {
+	for i := 1; i <= cap; i++ {
```
dbussink, line 43: Any reason for this change? Starting from `0` and using `<` is the canonical style.

internal/dumper/loader.go
```diff
@@ -62,37 +76,59 @@ func (l *Loader) Run(ctx context.Context) error {
...
+	// Commenting out to add some predictability to the order in which data files will be processed:
 	// Shuffle the tables
-	for i := range files.tables {
-		j := rand.Intn(i + 1)
-		files.tables[i], files.tables[j] = files.tables[j], files.tables[i]
-	}
+	//for i := range files.tables {
+	//	j := rand.Intn(i + 1)
+	//	files.tables[i], files.tables[j] = files.tables[j], files.tables[i]
+	//}
```
dbussink, line 94: I don't know if MySQL gives any guarantees about table ordering? So the start / end table logic is maybe not something very reliable to add. 

Otherwise we might need to change the query to show the tables so we can enforce the order there? That means changing it to use the `information_schema` tables then. 

In any way, if we remove code, we should remove it, not comment it :smile:. We have `git` to track things and find old code :wink:.

orware: I appreciate the input @dbussink ❤️!

Is your first sentence related to the `--starting-table` and `--ending-table` options? 

Those primarily help to drive the list of files that are being picked up from within the dump folder and mainly help provide some extra control over which files are going to be selected and used for the current `restore-dump` operation so the MySQL table ordering shouldn't necessarily factor in with those options?

I mainly see the `--starting-table`, `--ending-table` and several of the other new options added with this PR help ing to avoid needing to create multiple copies of a dump folder since normally you might have to create a new copy of an existing dump folder and then perform operations on it to keep just the tables / files you want to restore if you want to restore less than the full dataset.
---
As for the shuffling related code I commented out, the purpose of that code was initially confusing for me since I thought it did something different and I was unsure what the original rationale for adding it was.

I can definitely remove it completely from the file though and will make that correction shortly!

The shuffling only affects the order in which the data files get processed.

For example, with shuffling the data files will get processed out of order:

```
Started Processing Data File: metrics.connections.00005.sql in thread 3 (File 1 of 5)
Started Processing Data File: metrics.connections.00004.sql in thread 4 (File 2 of 5)
Started Processing Data File: metrics.connections.00001.sql in thread 1 (File 3 of 5)
Started Processing Data File: metrics.connections.00002.sql in thread 2 (File 4 of 5)
```

I only have 5 files in this example so it's not a great one but you can imagine things possibly starting at file 500, 1, 10, and 42 for a table with the shuffling option in place and then on a subsequent execution the data files would be processed again in a different order, etc.

I think removing the shuffling code is better for users since knowing that the files will be processed more or less in numeric order seems like a good thing? There is the caveat of threading still allowing for some processing out of order, but at least you can see that files 1-4 were started first below and then it will progress to file 5 afterward:

```
Started Processing Data File: metrics.connections.00001.sql in thread 3 (File 1 of 5)
Started Processing Data File: metrics.connections.00004.sql in thread 2 (File 4 of 5)
Started Processing Data File: metrics.connections.00003.sql in thread 1 (File 3 of 5)
Started Processing Data File: metrics.connections.00002.sql in thread 4 (File 2 of 5)
```

This seems like it would be useful for users since they would have some predictability for the order in which their data files will be processed by the `restore-dump` operation. 

This also comes out of working with a current user that is migrating from Postgres and generating files in the format needed for `restore-dump`.

Throughout that process different SQL errors have cropped up and it can be unclear which file needs to be reviewed more closely without the extra output that `--show-details` now provides. 

With the extra context of where you are in the processing of the files with `--show-details` this can help point a user in the right direction more quickly and they may even be able to more easily know where they can try and continue from in a subsequent `restore-dump` attempt without necessarily having to restore all of their data files again.

Going back to the shuffling example once more, if data file 500 for a table contained a problematic row when running a `restore-dump` attempt the amount of time it would take to encounter the error right now would vary since it would depend on where that file lands in the shuffled ordering. Without the shuffling we would find that the error should be encountered a fairly consistent manner since the data files will be processed in the same order.

internal/dumper/pool.go
orware: I've reverted this back to the canonical style in NewPool.

The only remaining question would be whether I should leave things as-is for `--show-details` with the zero-based output like this:

```
Started Processing Data File: metrics.connections.00001.sql in thread 2 (File 1 of 5)
Started Processing Data File: metrics.connections.00004.sql in thread 1 (File 4 of 5)
Started Processing Data File: metrics.connections.00002.sql in thread 3 (File 2 of 5)
Started Processing Data File: metrics.connections.00003.sql in thread 0 (File 3 of 5)
```

Or whether I should additionally tweak the output so it appears as 1-based.

If it's fine as-is then no additional updates should be needed.

internal/dumper/loader.go
orware: I removed the commented out shuffling code.

internal/dumper/loader.go
```diff
@@ -345,19 +441,47 @@ func (l *Loader) restoreTable(ctx context.Context, table string, conn *Connectio
 		return 0, err
 	}
 	query1 := string(data)
-	querys := strings.Split(query1, ";\n")
+	queries := strings.Split(query1, ";\n")
```
dbussink, line 444: Similar here, we should split with the parser instead.

internal/dumper/loader.go
```diff
@@ -214,8 +292,8 @@ func (l *Loader) restoreTableSchema(overwrite bool, tables []string, conn *Conne
 			return err
 		}
 		query1 := string(data)
-		querys := strings.Split(query1, ";\n")
-		for _, query := range querys {
+		queries := strings.Split(query1, ";\n")
```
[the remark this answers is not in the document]
orware: Incorporated these adjustments into this location and the 2 other locations where it made sense to add it as well.

internal/dumper/loader.go
orware: Incorporated these adjustments into the most recent commit, thank you @dbussink!

dbussink (approved):

Issue #916, opened 2024-10-07
handle homebrew's bump to mysql@v9 on macOS


aybabtme:
`pscale shell` is semi broken as of recently, because homebrew bumped mysql to v9, which doesnt contain the `.so` required for y'all's shell integration.

See https://stackoverflow.com/questions/78938322/mysql-authentication-plugin-issues-on-macos

https://gist.github.com/aybabtme/4fd51e181942405052d8e4645773a2d1

Might be good to catch that in the CLI and handle it cleanly, perhaps suggesting how to fix it at a minimum, so people don't wonder what's up.

dbussink:
Fixed!

PR #918, opened 2024-10-09
Fix the MySQL version detection and only use up to MySQL 8.x

dbussink:
MySQL 9.x is backwards compatibility breaking as it removes support for `mysql_native_password` which is what is used with PlanetScale as the default at the moment.

So enforce MySQL 8.x as the latest at the moment.

Also upgrade `planetscale-go` and add some empty test mocks for now so it can compile. This update removes a bunch of legacy code so it can't be used accidentally either.


dbussink:
Fixes #916

dbussink:
I had https://github.com/planetscale/psdbproxy/pull/24 as an alternative idea here that's more future proof but breaks some really old clients with the CLI / proxy.

iheanyi (approved):

dbussink:
Closing in favor of #919

PR #919, opened 2024-10-09
Switch shell / connect to use caching_sha2_password by default

dbussink:
This switches up both connect and shell to use `caching_sha2_password` by default. `connect` gets an option to also select the legacy `mysql_native_password` option. `shell` selects based on the `mysql --version` returns. For MySQL 8.x and later, it now uses `caching_sha2_password`.

Alternative to #918 

Depends on https://github.com/planetscale/psdbproxy/pull/24


iheanyi (approved):

dbussink:
Fixes #916

PR #950, opened 2024-12-13
Improve service token value input processing

orware:
Addresses a situation that can currently occur somewhat easily with the `--service-token-id` and `--service-token` values passed in via the CLI.

If the values are accidentally passed in a swapped manner, then when attempting to run your CLI command a user might receive something like the following error, rather than a more specific error message about the values being swapped:

```
Error: organization <ORG> does not exist or your account is not authorized to access it
```

There are a few ways to handle this situation, such as having a hard error thrown during the `IsAuthenticated()` check to direct users to the issue to automatically swapping the values internally for the user if it is detected they have provided values in the correct format that have simply been swapped.

For this pull request the more automatic approach has been used so if a swap is detected, the values will be swapped to the correct variables first and then used, which generally should lead to things simply working for users in these cases.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -86,7 +86,6 @@ func (c *Config) NewClientFromConfig(clientOpts ...ps.ClientOption) (*ps.Client,
 	}
 
 	if c.ServiceTokenIsSwapped() {
-		fmt.Println("The --service-token and --service-token-id are currently swapped.")
 		var correctServiceTokenID = c.ServiceToken
 		var correctServiceToken = c.ServiceTokenID
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -56,10 +56,6 @@ func (c *Config) IsAuthenticated() error {
 		return errors.New("both --service-token and --service-token-id are required for service token authentication")
 	}
 
-	if c.ServiceTokenIsSwapped() {
-		return errors.New("the --service-token and --service-token-id values are swapped")
-	}
-
 	if c.ServiceTokenIsSet() {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


dbussink:
Honestly I think we should do a better error message.

Especially since these are sensitive security credentials we shouldn't automatically switch them for people. I think for security things that risky and doesn't tell people which part needs to be kept secret and what isn't sensitive.

dbussink:
Compare it to a form where you login with a username and password, I don't think I've ever seen anything where if you swap those it still allows you to login.

orware:
@dbussink,

I adjusted things to use the stronger error approach suggested with the check within the `IsAuthenticated()` method.

Now if the values are provided in a swapped manner output like the following should be returned instead:

```
Error: not authenticated yet. Please run 'pscale auth login'
Error: the --service-token and --service-token-id values are swapped
```

internal/config/config.go
```diff
@@ -56,6 +56,10 @@ func (c *Config) IsAuthenticated() error {
 		return errors.New("both --service-token and --service-token-id are required for service token authentication")
 	}
 
+	if c.ServiceTokenIsSwapped() {
+		return errors.New("the --service-token and --service-token-id values are swapped")
```
dbussink, line 60: Wonder if it's nicer to say to the user "it looks like you swapped..."? No super strong opinions there though.

orware: How does `the --service-token and --service-token-id values appear to be swapped` sound to you?

orware: Updated the wording of the message to your example so it should look more like this now:

```
Error: not authenticated yet. Please run 'pscale auth login'
Error: it looks like you swapped the --service-token and --service-token-id values
```

dbussink (approved):

PR #968, opened 2025-02-17
Update to Go 1.24

dbussink:
Using the latest version of Go here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.23.2 as build
+FROM golang:1.24.0 as build
 WORKDIR /app
 COPY . .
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.23.2 as build
+FROM golang:1.24.0 as build
```
nickvanw, line 1: Should we pin this to `1.24` so that we get point releases automatically, or is that too risky for the CLI?

nickvanw (approved):

Dockerfile
GrahamCampbell: The other advantage of pinning with x.y.z is cache busting.

PR #1007, opened 2025-03-25
Add `keyspace settings` and `keyspace update-settings` commands

notfelineit:
This pull request implements a command for updating the keyspace advanced settings, in addition to implementing a command to show the current settings. The settings are isolated from `pscale keyspace show` because they are way too long in names to easily show in the table for it, so this is easier to work with.


internal/cmd/keyspace/settings.go
```diff
@@ -0,0 +1,89 @@
...
+	if ks.ReplicationDurabilityConstraints != nil {
+		// Map API values to more user-friendly display values
+		var constraint string
+		switch ks.ReplicationDurabilityConstraints.Strategy {
+		case "available":
+			constraint = "maximum"
+		case "always":
+			constraint = "minimum"
+		case "lag":
+			constraint = "dynamic"
```
notfelineit, line 70: nice!

dgraham (approved):

PR #1010, opened 2025-03-26
Add confirmation for potentially dangerous workflow actions

iheanyi:
There are some actions that can cause problems if done improperly. This adds a confirmation step for those actions.

### DIFF
--- a/internal/cmd/workflow/cancel.go
+++ b/internal/cmd/workflow/cancel.go
@@ -1,9 +1,13 @@
 package workflow
 
 import (
+	"errors"
 	"fmt"
+	"os"
 	"strconv"
 
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/planetscale/cli/internal/cmdutil"
 	"github.com/planetscale/cli/internal/printer"
 	ps "github.com/planetscale/planetscale-go/planetscale"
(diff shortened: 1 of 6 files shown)


internal/cmd/workflow/cancel.go
```diff
@@ -32,6 +38,35 @@ marks it as cancelled, allowing you to start a new workflow if needed.`,
...
+					return fmt.Errorf("cannot cancel workflow with the output format %q (run with -force to override)", ch.Printer.Format())
+				}
+
+				if !printer.IsTTY {
+					return fmt.Errorf("cannot confirm cancellation (run with -force to override)")
+				}
+
+				prompt := &survey.Confirm{
+					Message: "Are you sure you want to cancel this workflow?",
+					Default: false,
```
notfelineit, line 52: ❤️

notfelineit (approved):

internal/cmd/workflow/cutover.go
```diff
@@ -61,5 +112,7 @@ func CutoverCmd(ch *cmdutil.Helper) *cobra.Command {
 		},
 	}
 
+	cmd.Flags().BoolVar(&force, "force", false, "Force thecutover without prompting for confirmation.")
```
GrahamCampbell, line 115: typo in "thecutover"

iheanyi: good catch, thank you

PR #1046, opened 2025-04-30
Fix branch create help message

jgreet:
I'm not sure why but the backticks change the description of the `-cluster-size` flag in a confusing way.

On 0.241.0 

```
$ pscale branch create
Error: missing arguments <source-database, branch>

Usage:
  pscale branch create <source-database> <branch> [options] [flags]

Aliases:
  create, b

Flags:
      --cluster-size pscale size cluster list   Cluster size for branches being created from a backup or seeded with data. Use pscale size cluster list to see the valid sizes. (default "PS-10")
      --from string                             Parent branch to create the new branch from. Cannot be used with --restore
...
```

### DIFF
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -106,9 +106,9 @@ func ApiCmd(ch *cmdutil.Helper, userAgent string, defaultHeaders map[string]stri
 
 	opts := &ApiOpts{}
 	cmd.Flags().StringVarP(&opts.Method, "method", "X", "GET", "HTTP method to use for the request.  Defaults to GET for requests without a body, or POST when a body is specified with --field/-F or --input/-I")
-	cmd.Flags().StringArrayVarP(&opts.Query, "query", "Q", nil, "query to append to the URL path, in key=value format")
+	cmd.Flags().StringArrayVarP(&opts.Query, "query", "Q", nil, "query to append to the URL path, in `key=value` format")
 	cmd.Flags().StringArrayVarP(&opts.Header, "header", "H", nil, "HTTP headers to add to the request")
-	cmd.Flags().StringArrayVarP(&opts.Field, "field", "F", nil, "HTTP body to send with the request, in key=value format where value is a JSON entity, unless value starts with a @ in which case the string after @ represents a file that will be read. Nested types are represented as 'root.depth1.depth2=value'")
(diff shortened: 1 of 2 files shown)


iheanyi (approved):
Can we check other uses of this flag and update their help text too?

jgreet:
Yeah, taking a look now. There's at least one more in `pscale shell`

jgreet:
Actually in `shell --help` it works. May even be intentional. Same in a few other places

https://github.com/planetscale/cli/blob/581cc298c29e838c21a47cb9599f18648838b8f9/internal/cmd/shell/shell.go#L240
```
      --remote-addr hostname:port   PlanetScale Database remote network address. By default the remote address is populated automatically from the PlanetScale API. (format: hostname:port)
```
---
But `api --help` looks wrong, see `--method` or the two backticks at the end of `--field`

before:
```
  -F, --field key=value      HTTP body to send with the request, in key=value format where `value` is a JSON entity, unless `value` starts with a `@` in which case the string after `@` represents a file that will be read. Nested types are represented as `root.depth1.depth2=value``
  -H, --header stringArray   HTTP headers to add to the request
  -h, --help                 help for api
  -I, --input string         HTTP body to send with the request, as a file that will be read and then sent.
  -X, --method --field/-F    HTTP method to use for the request.  Defaults to GET for requests without a body, or POST when a body is specified with --field/-F or `--input/-I` (default "GET")
      --org string           The organization for the current user
  -Q, --query key=value      query to append to the URL path, in key=value format
```

after:
```
Flags:
      --branch string        The branch this project is using
      --database string      The database this project is using
  -F, --field stringArray    HTTP body to send with the request, in key=value format where value is a JSON entity, unless value starts with a @ in which case the string after @ represents a file that will be read. Nested types are represented as 'root.depth1.depth2=value'
  -H, --header stringArray   HTTP headers to add to the request
  -h, --help                 help for api
  -I, --input string         HTTP body to send with the request, as a file that will be read and then sent.
  -X, --method string        HTTP method to use for the request.  Defaults to GET for requests without a body, or POST when a body is specified with --field/-F or --input/-I (default "GET")
      --org string           The organization for the current user
  -Q, --query stringArray    query to append to the URL path, in key=value format
```

jgreet:
Here's what's happening

https://github.com/spf13/pflag/blob/d661846b4df0d6611ad95577ddfb240474d21b7c/flag.go#L566-L570

PR #1147, opened 2025-11-22
add flag to install pscale MCP to zed as well

Shubhaankar-Sharma:
This PR abstracts the mcp installation code a bit as zed requires MCP servers to also live in their main settings.json

their JSON files are JSONC (with comments) thus I had to use a jsonc library to parse it


internal/cmd/mcp/zed.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return nil
+}
+
+func storeContextServers(configPath string, cfg *zedConfig) error {
+	var fullSettings map[string]any
+
+	if fileData, err := os.ReadFile(configPath); err == nil {
+		backupPath := configPath + ".backup"
```
piki:
```suggestion
		backupPath := configPath + "~"
```

piki (approved):
Looks like it will work.  Have you tested it with an actual copy of Zed, to make sure the Zed Agent can use the PlanetScale mcp?

The implementations for Claude Desktop, Cursor, and Zed are redundant and all a bit different here.  If you're into it, it could use a refactor so there's a single function that handles things like creating the containing directory, saving a backup file, and adding our config to the overall JSONC blob.  (You can parse all three files as JSONC, even though we expect Claude and Cursor not to have any comments.)  It could take the directory, the basename of the JSON config file, and a function parameter that modifies the deserialized JSON object to add the right config to it.

internal/cmd/mcp/zed.go
```diff
@@ -0,0 +1,108 @@
...
+		if err := os.WriteFile(backupPath, fileData, 0644); err != nil {
```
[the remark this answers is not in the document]
Shubhaankar-Sharma: got it, thanks!

internal/cmd/mcp/zed.go
Shubhaankar-Sharma: made this change

PR #1287, opened 2026-07-08
Support Postgres branches in branch infra

nickvanw:
## Summary

`pscale branch infra` only worked for Vitess branches. The branch infrastructure API now serves Postgres branches too, so this bumps planetscale-go to [v0.177.0](https://github.com/planetscale/planetscale-go/releases/tag/v0.177.0) and renders the Postgres response.

## Changes

- Bump `planetscale-go` v0.176.0 → v0.177.0, which models `BranchInfrastructure` as a per-engine union (`Vitess` / `Postgres`); the Vitess path is updated for the new field layout with no behavior change.
- For Postgres branches, render a Postgres-native table: one row per node (`postgres`) and per bouncer (`pgbouncer`) with role, size, availability zone, and storage usage/capacity. JSON/CSV output marshals the underlying SDK objects, matching how the Vitess pods table behaves.

## Why

The API returns an engine-native shape per database kind rather than forcing Postgres data into the Vitess pods shape, so the command branches once on which engine the response describes. The command remains hidden and the API is feature-gated with no stability guarantee.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/charmbracelet/huh v1.0.0
 	github.com/charmbracelet/lipgloss v1.1.0
 	github.com/charmbracelet/x/ansi v0.11.2
+	github.com/dustin/go-humanize v1.0.1
 	github.com/fatih/color v1.19.0
 	github.com/frankban/quicktest v1.14.6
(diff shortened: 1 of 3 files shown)


dgraham (approved):
⚡️

PR #1298, opened 2026-07-20
Add Postgres parameter management to branch resize, plus change status/cancel

no-itsbackpack:
## Summary

Adds first-class support for managing Postgres configuration parameters from the CLI, built on the branch changes API. Cluster size, replicas, and parameters are all upserted into a single asynchronous change request, so one `resize` invocation can change any combination atomically (one maintenance operation, one restart). Also adds visibility and control over the change request lifecycle, which `resize` previously lacked entirely.

## New commands and flags

```bash
# Read the parameter catalog (names, current/default values, restart & immutable flags)
pscale branch parameters list <database> <branch> [--namespace pgconf] [--extension] [--internal]

# Change parameters — repeatable flag, keys are namespace.name
pscale branch resize <database> <branch> --parameters pgconf.max_connections=200

# Combine size, replicas, and parameters into one change request
pscale branch resize <database> <branch> --cluster-size PS_20_GCP_X86 --replicas 2 \
  --parameters pgconf.max_connections=500 --wait

# Track / cancel the change request
pscale branch resize status <database> <branch>
pscale branch resize cancel <database> <branch>
```

- `--parameters` values are preflighted against the catalog: unknown or immutable parameters fail fast without creating a change request, and restart-requiring parameters are called out.
- `--wait` polls until the change reaches a terminal state (`completed`/`canceled`), with `--wait-timeout` (default 10m). Transient poll errors retry, matching the other `--wait` commands.
- `resize` now requires at least one of `--cluster-size`, `--replicas`, `--parameters`.

## Agent/automation friendliness

- All commands emit full API objects with `--format json`; a no-op change prints `{"result": "no_change"}` on stdout instead of a stderr-only note.
- Parameter validation failures from the API used a nested error shape the client didn't parse, so users saw `internal error, response body doesn't match error type signature`. That shape is now parsed into the real message (e.g. `pgconf.max_connections: Value must be less than or equal to 100`), and the generic fallback errors now include the response body so no unrecognized shape can hide the cause again.
- `AGENTS.md` (embedded agent guide) documents the workflow, change states, and JSON result shapes.

## Test plan

- [x] Unit tests: httptest client tests for all new endpoints (including the bare-array parameters response and string-typed min/max), command tests for preflight rejection, no-op JSON output, MySQL redirect, and `--parameters` parsing edge cases
- [x] `script/test-parameters.sh` (new) sweeps every parameter in a branch's catalog against local api-bb: 292 checks across 155 parameters — preflight rejections, no-op round-trips, out-of-range and invalid-option validation messages, and a full change/status/cancel lifecycle with value restoration. All passing.
- [x] `go vet`, staticcheck, gofmt clean


joshmgross (approved):

script/test-parameters.sh
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env bash
```
no-itsbackpack: will be removed

PR #1365, opened 2026-08-20
Add pscale --skill to print an installable agent skill file

nickvanw:
## What

Adds `pscale --skill`, mirroring [`herdr --skill`](https://herdr.dev): a cold-start one-liner that prints the agent guide as an **installable skill file** (YAML frontmatter trigger + the embedded guide), so an agent with nothing but the binary can drop it into a skills directory and get trigger-based loading.

`pscale --skill` is an alias for `pscale agent-guide --skill`; both render from the same embedded `AGENTS.md`, so there is no content fork.

## Changes

- **`internal/cmd/root.go`** — new root `--skill` flag. Because cobra prints root help (never a PreRun) when no subcommand is given, it scans `os.Args` before `Execute`, exactly like `herdr --skill`. Root `Long` and the `--help` footer now lead with `pscale --skill`.
- **`internal/cmd/agentguide/agentguide.go`** — `SkillDoc()` prepends `name`/`description` frontmatter to the embedded guide. `agent-guide` (bare) still prints the plain guide for humans; `agent-guide --skill` prints the skill file.
- **`AGENTS.md`** — skill-ified:
  - New **Concepts** section: org → database → branch hierarchy, engine disambiguation (deploy-request / password / keyspace resize = Vitess; role / branch resize / import d1 = Postgres), and a "weird database" triage (insights / inspect / metrics).
  - Removed the repo-dev note (moved to README) — meaningless inside an installed skill.
  - De-circularized self-references: "start here" and Workflow step 1 now point at `pscale --skill`.
- **`README.md`** — the repo-dev pointer (`internal/planetscale/`, `doc/api-client.md`) is preserved under Local Development; Docs section mentions both entry points.
- **Tests** — `TestSkillDoc` (frontmatter + body present), `TestAgentGuideSkillFlag` (`--skill` emits frontmatter).

## Verification

- `pscale --skill` ≡ `pscale agent-guide --skill` (byte-identical, exit 0); `agent-guide` bare stays frontmatter-free; `--help` and other commands unaffected.
- Local CI-equivalent all green: `go fmt`, `go vet`, `staticcheck`, `golangci-lint` (0 issues), full `go test ./...`.

## Note

The `--skill` `os.Args` scan triggers on any bare `--skill` token before cobra runs. No real command defines a `--skill` flag, so this is harmless, but it can be scoped to position 1 if preferred.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,15 +1,11 @@
 # PlanetScale CLI — agent guide
 
-> **Developing the CLI?** The API client is vendored at `internal/planetscale/`;
-> this repo **no longer depends on `planetscale-go`**. Read `doc/api-client.md`
-> before touching API-facing code. The rest of this file is about *using* `pscale`.
-
 For **any** automated agent or script using `pscale`. Always pass **`--format json`**. Substitute placeholders from the user's request or from prior command output (`org list`, `database list`, `branch list`).
 
-If you only have the installed `pscale` binary, start here:
+If you only have the installed `pscale` binary and this guide is not already in your context, load it and check auth:
 
 ```bash
-pscale agent-guide --format json
+pscale --skill
 pscale auth check --format json
 ```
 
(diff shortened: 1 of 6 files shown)


nickvanw:
Addressed the Bugbot finding in 8dd58850 — it was correct.

**Root cause of my earlier confusion:** the root `os.Args` scan was intercepting the exact token `--skill` and printing via `fmt.Print`, which *masked* the subcommand bug during my testing. The true subcommand path (`agent-guide --skill=true --format json`, which the scan misses) did hit `Printer.Print` → `io.Discard` → exit 0 with empty stdout, exactly as reported.

**Fixes:**
1. `agent-guide --skill` now uses `fmt.Print` (a raw file dump, like `--version`), bypassing the format-aware printer — verified non-empty in both human and JSON modes.
2. Replaced the `os.Args` scan with a root `RunE` + a normal bool flag (the idiomatic cobra pattern, same mechanism as `--version`). This also fixes `pscale --skill=true` falling through to root help.
3. `TestAgentGuideSkillFlag` now captures `os.Stdout` and asserts output in both human and JSON formats; the prior `SetHumanOutput` capture was masking the discard path.

Local CI-equivalent green: `go fmt`, `go vet`, `staticcheck`, `golangci-lint` (0 issues), full `go test ./...`.

nholden (approved):

PR #1404, opened 2026-09-10
Add customer keyspace rollout concurrency settings

darwich6:
Customers cannot currently configure the existing keyspace shard rollout concurrency from the CLI. Add `pscale keyspace update-settings --max-rollout 1..32` and `--reset-max-rollout`, with readback through `keyspace settings`.

The vendored API client preserves the difference between an omitted setting, an explicit integer, and JSON `null` for reset. Human output shows the configured integer, defaulting to 1 when unset; JSON returns the stored nullable value. Settings PATCHes now include only selected groups, avoiding unrelated durability updates on Metal while preserving unspecified flags within VReplication updates. The new flags reject interactive-mode combinations.

Validation:
- Full `go test ./...`, `go build -trimpath ./...`, and `go vet ./...` passed.
- Formatting and `git diff --check` passed.
- Local CLI → running development API smoke test passed for set/read/reset and preserving unrelated settings; original local value restored. Direct HTTP validation also rejected 33 with HTTP 422.
- GitHub CI passed: Build, Test, & Lint; lint (including golangci-lint); and dependency license verification.

### DIFF
--- a/internal/cmd/keyspace/settings.go
+++ b/internal/cmd/keyspace/settings.go
@@ -53,7 +53,11 @@ func SettingsCmd(ch *cmdutil.Helper) *cobra.Command {
 // toKeyspaceSettings converts a Keyspace API response to a KeyspaceSettings object for display
 func toKeyspaceSettings(ks *ps.Keyspace) *KeyspaceSettings {
 	settings := &KeyspaceSettings{
-		orig: ks,
+		MaxRollout: 1,
+		orig:       ks,
+	}
+	if ks.MaxRollout != nil {
+		settings.MaxRollout = *ks.MaxRollout
 	}
 
 	// Set replication durability constraints if available
--- a/internal/cmd/keyspace/settings_test.go
+++ b/internal/cmd/keyspace/settings_test.go
@@ -3,6 +3,7 @@ package keyspace
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"testing"
 	"time"
@@ -183,6 +184,7 @@ func TestBuildKeyspaceSettings(t *testing.T) {
 	c := qt.New(t)
 
 	ts := time.Now()
+	maxRollout := 64
 
 	// Test with all settings populated
 	fullKs := &ps.Keyspace{
@@ -198,6 +200,7 @@ func TestBuildKeyspaceSettings(t *testing.T) {
 			AllowNoBlobBinlogRowImage: true,
 			VPlayerBatching:           false,
 		},
+		MaxRollout: &maxRollout,
 	}
 
 	settings := toKeyspaceSettings(fullKs)
@@ -205,6 +208,8 @@ func TestBuildKeyspaceSettings(t *testing.T) {
 	c.Assert(settings.VReplicationFlags.OptimizeInserts, qt.Equals, true)
 	c.Assert(settings.VReplicationFlags.AllowNoBlobBinlogRowImage, qt.Equals, true)
 	c.Assert(settings.VReplicationFlags.VPlayerBatching, qt.Equals, false)
+	c.Assert(settings.MaxRollout, qt.Equals, 64)
+	assertMaxRolloutJSON(t, settings, "64")
 
 	// Test with nil settings
 	nilKs := &ps.Keyspace{
@@ -221,4 +226,17 @@ func TestBuildKeyspaceSettings(t *testing.T) {
 	c.Assert(nilSettings.VReplicationFlags.OptimizeInserts, qt.Equals, false) // Default values
 	c.Assert(nilSettings.VReplicationFlags.AllowNoBlobBinlogRowImage, qt.Equals, false)
 	c.Assert(nilSettings.VReplicationFlags.VPlayerBatching, qt.Equals, false)
+	c.Assert(nilSettings.MaxRollout, qt.Equals, 1)
+	assertMaxRolloutJSON(t, nilSettings, "null")
+}
+
+func assertMaxRolloutJSON(t *testing.T, settings *KeyspaceSettings, want string) {
+	t.Helper()
+	c := qt.New(t)
+	encoded, err := json.Marshal(settings)
+	c.Assert(err, qt.IsNil)
+
+	var object map[string]json.RawMessage
+	c.Assert(json.Unmarshal(encoded, &object), qt.IsNil)
+	c.Assert(string(object["max_rollout"]), qt.Equals, want)
 }
--- a/internal/cmd/keyspace/update_settings.go
+++ b/internal/cmd/keyspace/update_settings.go
@@ -12,11 +12,11 @@ import (
 )
 
 func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
-	updateReq := &ps.UpdateKeyspaceSettingsRequest{}
-
 	var flags struct {
 		replicationDurabilityConstraints *ps.ReplicationDurabilityConstraints
 		vreplicationFlags                *ps.VReplicationFlags
+		maxRollout                       int
+		resetMaxRollout                  bool
 		interactive                      bool
 	}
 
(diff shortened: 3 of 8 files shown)


internal/cmd/keyspace/update_settings.go
```diff
@@ -129,19 +151,16 @@ func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
 	cmd.Flags().BoolVar(&flags.vreplicationFlags.OptimizeInserts, "vreplication-optimize-inserts", true, "When enabled, skips sending INSERT events for rows that have yet to be replicated.")
 	cmd.Flags().BoolVar(&flags.vreplicationFlags.AllowNoBlobBinlogRowImage, "vreplication-enable-noblob-binlog-mode", true, "When enabled, omits changed BLOB and TEXT columns from replication events, which reduces binlog sizes.")
 	cmd.Flags().BoolVar(&flags.vreplicationFlags.VPlayerBatching, "vreplication-batch-replication-events", false, "When enabled, sends fewer queries to MySQL to improve performance.")
+	cmd.Flags().IntVar(&flags.maxRollout, "max-rollout", 1, "Maximum number of concurrent shard rollouts (1-32). The effective service cap is 32.")
+	cmd.Flags().BoolVar(&flags.resetMaxRollout, "reset-max-rollout", false, "Reset the configured maximum concurrent shard rollouts to the default (1).")
```
maxenglander: i don't think we use this pattern elsewhere do we? if not i would not introduce this pattern. i would update `max-rollout` to document the default as `1`, users can manually reset by passing `--max-rollout=1`

internal/cmd/keyspace/settings_test.go
```diff
@@ -221,4 +226,17 @@ func TestBuildKeyspaceSettings(t *testing.T) {
 	c.Assert(nilSettings.VReplicationFlags.OptimizeInserts, qt.Equals, false) // Default values
 	c.Assert(nilSettings.VReplicationFlags.AllowNoBlobBinlogRowImage, qt.Equals, false)
 	c.Assert(nilSettings.VReplicationFlags.VPlayerBatching, qt.Equals, false)
+	c.Assert(nilSettings.MaxRollout, qt.Equals, 1)
+	assertMaxRolloutJSON(t, nilSettings, "null")
```
maxenglander: unclear to me why we need this..are we doing this for other numeric flags?

internal/cmd/keyspace/update_settings.go
```diff
@@ -34,16 +34,48 @@ func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
...
 			database, branch, keyspace := args[0], args[1], args[2]
+			maxRolloutChanged := cmd.Flags().Changed("max-rollout")
+			resetMaxRolloutChanged := cmd.Flags().Changed("reset-max-rollout")
+			resetMaxRolloutRequested := resetMaxRolloutChanged && flags.resetMaxRollout
+
+			if maxRolloutChanged && resetMaxRolloutRequested {
+				return fmt.Errorf("--max-rollout and --reset-max-rollout are mutually exclusive")
+			}
+			if flags.interactive && (maxRolloutChanged || resetMaxRolloutChanged) {
+				return fmt.Errorf("--max-rollout and --reset-max-rollout cannot be used with --interactive")
```
maxenglander: are we doing this exclusion logic for other flags? if not, i would defer this to a separate PR that applies this guard for all flags.

internal/cmd/keyspace/update_settings.go
```diff
@@ -34,16 +34,48 @@ func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
...
+			}
+			if maxRolloutChanged && (flags.maxRollout < 1 || flags.maxRollout > 32) {
+				return fmt.Errorf("--max-rollout must be between 1 and 32")
+			}
+			if cmd.Flags().Changed("throttler-threshold") && flags.throttlerThreshold < 0 {
+				return errors.New("--throttler-threshold must be greater than or equal to 0")
+			}
```
maxenglander: would recommend handling changes to unrelated flags in a separate PR to keep this one tightly focused

maxenglander (commented):
Thanks for putting this together. I'd like to scope this PR down to the minimal change needed to support `--max-rollout`, following the patterns already in `update-settings` (the throttler flags are the closest precedent), in a way that doesn't change or put at risk any existing functionality.

Concretely, I'd like this PR to contain only:

- `MaxRollout *int` on `Keyspace` and on `UpdateKeyspaceSettingsRequest` (with `omitempty`), like `Throttler.Threshold`.
- An `IntVar` flag for `--max-rollout`, set on the request when `cmd.Flags().Changed("max-rollout")`, and included in the "no changes were requested" guard.
- Displaying `max rollout` in `keyspace settings` the same way sibling nullable fields are displayed (`"not set"` when the API returns null).
- Tests for the above.

Everything else here is worthwhile but separable, and I'd prefer to see each in its own PR so it can be reviewed and reverted on its own:

- **Skipping the GET / sending only changed setting groups.** This changes the PATCH payload shape for all existing flags, flips several pre-existing test assertions, and changes the exit status of a flagless invocation against a nonexistent keyspace. It also isn't applied to `--interactive`, so the Metal problem described in the PR body still exists there. This deserves its own PR with its own description.
- **`--reset-max-rollout` and the `**int` tri-state.** I would like to avoid this option entirely. I think `--max-rollout` should be the only way to change max. rollout.
- **Hand-rolled flag exclusivity checks.** If we keep multiple flags that conflict, `cmd.MarkFlagsMutuallyExclusive` is what the rest of the repo uses (see `webhook/update.go`).
- **Hardcoding the default of `1` in the display layer.** Siblings render null as `"not set"`; baking the server default into the client means we show a wrong value if the default ever changes.

Here's roughly what I have in mind for the minimal version (against `main`, untested, tests omitted):

```diff
diff --git a/internal/planetscale/keyspaces.go b/internal/planetscale/keyspaces.go
--- a/internal/planetscale/keyspaces.go
+++ b/internal/planetscale/keyspaces.go
@@ type Keyspace struct {
 	VReplicationFlags                *VReplicationFlags                `json:"vreplication_flags"`
 	ReplicationDurabilityConstraints *ReplicationDurabilityConstraints `json:"replication_durability_constraints"`
+	MaxRollout                       *int                              `json:"max_rollout"`
 	Throttler                        *KeyspaceThrottler                `json:"throttler"`
 	ReadOnlyRegions                  []*ReadOnlyRegionKeyspace         `json:"read_only_regions"`
 }
@@ type UpdateKeyspaceSettingsRequest struct {
 	ReplicationDurabilityConstraints *ReplicationDurabilityConstraints `json:"replication_durability_constraints,omitempty"`
 	VReplicationFlags                *VReplicationFlags                `json:"vreplication_flags,omitempty"`
 	Throttler                        *KeyspaceThrottler                `json:"throttler,omitempty"`
+	MaxRollout                       *int                              `json:"max_rollout,omitempty"`
 }
diff --git a/internal/cmd/keyspace/update_settings.go b/internal/cmd/keyspace/update_settings.go
--- a/internal/cmd/keyspace/update_settings.go
+++ b/internal/cmd/keyspace/update_settings.go
@@ func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
 		throttlerEnabled                 bool
 		throttlerThreshold               float64
+		maxRollout                       int
 		interactive                      bool
 	}
@@ func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
 					updateReq.Throttler.Threshold = &flags.throttlerThreshold
 				}
 			}
 
-			if !rdcChanged && !vrfChanged && !throttlerChanged {
+			maxRolloutChanged := cmd.Flags().Changed("max-rollout")
+			if maxRolloutChanged {
+				if flags.maxRollout < 1 || flags.maxRollout > 32 {
+					return errors.New("--max-rollout must be between 1 and 32")
+				}
+				updateReq.MaxRollout = &flags.maxRollout
+			}
+
+			if !rdcChanged && !vrfChanged && !throttlerChanged && !maxRolloutChanged {
 				end()
 				ch.Printer.Println("No changes were requested. No update performed.")
 				return nil
 			}
@@ func UpdateSettingsCmd(ch *cmdutil.Helper) *cobra.Command {
 	cmd.Flags().Float64Var(&flags.throttlerThreshold, "throttler-threshold", 5, "Replication lag in seconds above which migrations and workflows are paused.")
+	cmd.Flags().IntVar(&flags.maxRollout, "max-rollout", 1, "Maximum number of shards to roll out changes to concurrently (1-32).")
 	cmd.Flags().BoolVarP(&flags.interactive, "interactive", "i", false, "Run the command in interactive mode")
diff --git a/internal/cmd/keyspace/keyspace.go b/internal/cmd/keyspace/keyspace.go
--- a/internal/cmd/keyspace/keyspace.go
+++ b/internal/cmd/keyspace/keyspace.go
@@ type KeyspaceSettings struct {
 	ReplicationDurabilityConstraintStrategy string            `header:"replication durability constraint strategy" json:"replication_durability_constraint"`
 	VReplicationFlags                       VReplicationFlags `header:"inline" json:"vreplication_flags"`
+	MaxRollout                              string            `header:"max rollout" json:"max_rollout"`
 	Throttler                               Throttler         `header:"inline" json:"throttler"`
diff --git a/internal/cmd/keyspace/settings.go b/internal/cmd/keyspace/settings.go
--- a/internal/cmd/keyspace/settings.go
+++ b/internal/cmd/keyspace/settings.go
@@ func toKeyspaceSettings(ks *ps.Keyspace) *KeyspaceSettings {
 	settings := &KeyspaceSettings{
-		orig: ks,
+		MaxRollout: "not set",
+		orig:       ks,
 	}
+	if ks.MaxRollout != nil {
+		settings.MaxRollout = strconv.Itoa(*ks.MaxRollout)
+	}
```

Happy to talk through any of this if I'm missing context on why the broader changes need to land together.

internal/cmd/keyspace/settings.go
```diff
@@ -53,7 +54,12 @@ func SettingsCmd(ch *cmdutil.Helper) *cobra.Command {
 // toKeyspaceSettings converts a Keyspace API response to a KeyspaceSettings object for display
 func toKeyspaceSettings(ks *ps.Keyspace) *KeyspaceSettings {
 	settings := &KeyspaceSettings{
-		orig: ks,
+		MaxRollout: "not set",
+		orig:       ks,
+	}
+
+	if ks.MaxRollout != nil {
```
maxenglander, line 61: the API should always come back with a response right? i think "not set" is kind of misleading. a value _is_ always set to a value between 1-32.

maxenglander (approved):
one question but otherwise lgtm

mscoutermarsh (approved):

